PR #2, opened 2016-09-09
Update global.json

ewinnington:
Updating the version of the platform targeted, which I think would solve the Nuget issue.

### DIFF
--- a/global.json
+++ b/global.json
@@ -1,6 +1,6 @@
 ﻿{
   "projects": [ "src", "test" ],
   "sdk": {
-    "version": "1.0.0-preview1-002702"
+    "version": "1.0.0-preview2-003121"
   }
 }


dmetzgar:
What's the nuget issue? global.json is not particularly useful and the .NET CLI doesn't build from it directly.

ewinnington:
For the Nuget, I get an error message when I try to install it into my core-minimal test environment:

Retrieving package 'Microsoft.CoreWf 1.0.0' from 'nuget.org'.
Installing NuGet package Microsoft.CoreWf.1.0.0.
Successfully installed 'Microsoft.CoreWf 1.0.0' to core-minimal
Executing nuget actions took 26.76 ms
========== Finished ==========
Time Elapsed: 00:00:00.3039885
PATH=.\node_modules.bin;C:\Program Files (x86)\Microsoft Visual Studio 14.0\Web\External;%PATH%;C:\Program Files (x86)\Microsoft Visual Studio 14.0\Web\External\git
C:\Program Files\dotnet\dotnet.exe restore "C:\Data-Dump\projects\core-minimal\src.vs\restore.dg"
log : Restoring packages for C:\Data-Dump\projects\core-minimal\src\project.json...
error: Package Microsoft.CoreWf 1.0.0 is not compatible with netcoreapp1.0 (.NETCoreApp,Version=v1.0). Package Microsoft.CoreWf 1.0.0 supports: net (.NETFramework,Version=v0.0)
error: One or more packages are incompatible with .NETCoreApp,Version=v1.0.
log : Writing lock file to disk. Path: C:\Data-Dump\projects\core-minimal\src\project.lock.json
log : C:\Data-Dump\projects\core-minimal\src\core-minimal.xproj
log : Restore failed in 1169ms.
Errors in C:\Data-Dump\projects\core-minimal\src\core-minimal.xproj
Package Microsoft.CoreWf 1.0.0 is not compatible with netcoreapp1.0 (.NETCoreApp,Version=v1.0). Package Microsoft.CoreWf 1.0.0 supports: net (.NETFramework,Version=v0.0)
One or more packages are incompatible with .NETCoreApp,Version=v1.0.

That went away when I generated a local nuget with the new target framework. I'm still struggling with the .net core compatibility.

ewinnington:
To reproduce, you could clone https://github.com/ewinnington/core-minimal and add the corewf to the project if you want.

dmetzgar:
Implementation of netstandard is not as good as it should be. I could try adding "netcoreapp" as a target framework. Also, the version specification in global.json should be completely removed.

dmetzgar:
I was able to add the reference and do a restore and build without issue. Can you verify that you have the latest .NET Core SDK? The new ones don't always replace the old ones so you could try explicitly uninstalling.

ewinnington:
Yeah something is fishy on my side. I'm uninstalling and resetting my core clr installs.

Issue #4, opened 2016-11-27
Automatic CacheMetadata


dmetzgar:
This should either be replaced with a reflection-based implementation or use the System.ComponentModel pieces from the .NET Standard 2.0.

ewinnington:
Is there a battery of tests that check the functionality of CacheMetadata? 

I un-commented the Automatic CacheMetadata in my [branch](https://github.com/ewinnington/corewf) after moving to net standard 2.0 and was able to run the simplest test of removing the implementation of CacheMetadata in the Samples/Sequential. That test passed successfully.

dmetzgar:
I think so. I'll add an issue to port some of the other tests we didn't port earlier.

dmetzgar:
Added #19

PR #22, opened 2018-01-05
Adding automatic cachemetadata by supporting .NET Standard 1.5

dmetzgar:
Need to add some tests before merging. There may be some components that I missed.

Looks like there is enough support for componentmodel in .NET Standard 1.5 to allow automatic cachemetadata to work. The csproj change builds both 1.3 and 1.5. The NuGet package would have both target frameworks in it.


dmetzgar:
Solution to issue #4: Automatic CacheMetadata

dmetzgar:
Closing this in favor of the other PR

Issue #16, opened 2017-07-19
Use Microsoft.Extensions.Localization instead of resxcodefilegenerator


dmetzgar:
In the traditional build path in the .NET Framework, we have a tool called resxcodefilegenerator that turns a resx file into a cs file. Developers can then reference use properties to reference the various resource strings instead of calling ResourceManager directly. A new library introduced with .NET Core called Microsoft.Extensions.Localization includes the IStringLocalizer. The default language can be used as keys directly in the code since it will use the key if there is no resource string to be found for that culture. This also makes it easier to find where a particular error message comes from in the code.

InariTheFox:
I wouldn't mind tackling this puppy and some other pieces around localization.

dmetzgar:
@InariTheFox, that would be great. I have to port the current localized resx files over to restext. I'll try first with the en-us version and check it in.

dmetzgar:
@InariTheFox, please take a look at #21 when you get a chance. This has the restext files for a bunch of languages. I'm kinda torn on whether the IStringLocalizer would be helpful. I'm not entirely satisfied with the SR class and how many lines it has just to do parameter substitution in the strings. It would be a lot of work to remove this SR class and I think it would depend on needing to add the IStringLocalizer to dependency injection and DI isn't initialized anywhere right now.

InariTheFox:
Thanks. I'll take a stab at it this weekend!

Issue #14, opened 2017-07-15
Add Transactions support


dmetzgar:
System.Transactions is part of .NET Core 2.0. It only supports local transactions, but that should be enough for some workflow scenarios.

ewinnington:
I started porting the transactions, but realized it was larger than I thought and needed some serious effort. Also, it will be much easier when we have Xaml support to generate transacted workflows. I'll get back to it when Xaml works on core and/or framework. 

Also, the tests will be vital here.

Issue #6, opened 2017-07-14
Use Portable.Xaml to implement the System.Activities.XamlIntegration


dmetzgar:
This is referring to @cwensley's port of Portable.Xaml to the .NET Standard here: https://github.com/cwensley/Portable.Xaml

It seems the most promising choice in terms of implementing the features of XAML that WF needs.

ewinnington:
I got the PortableXAML version compiling and running until I'm trying to create the Type and I get an exception: when the underlying type on a XamlType is known, but it is unknown for '{http://schemas.microsoft.com/netfx/2009/xaml/activities}Activity'' from the Portable.Xaml.dll!Portable.Xaml.Schema.XamlTypeInvoker.ThrowIfUnknown() Line 76 when it's trying to Create the instance of Activity (CreateInstance). 

Any idea?

Now I need to figure out how to register the types correctly... Pushing it to my branch as soon as it's cleaned up.

cwensley:
@ewinnington you need to add the Portable.Xaml.Markup.XmlnsDefinitionAttrbute to the assemblies that define those types.

E.g.  `[assembly: XmlnsDefinition("http://schemas.microsoft.com/netfx/2009/xaml/activities", "Clr.Namespace.To.Activity.In.This.Assembly")]`

You can also define these in another assembly by setting the AssemblyName property of the XmlnsDefinitionAttribute.

Note that if you are instantiating your own `XamlSchemaContext` (which I recommend using a static instance if performance is a concern), that the assemblies you pass in the constructor must contain all the types you wish to use in xaml.

Hope this helps!

ewinnington:
Wow! Thanks for the information. I'll do that and report the effect :)
You saved me a ton of time searching.

Update: Didn't seem to do the trick. I'll clean up the commit, push to my branch and have a tea. 
  
I'm not instantiating my XamlSchemaContext. My code is: 

```
string XamlText = @"
<Activity x:Class=""WFTemplate"" 
mva:VisualBasic.Settings=""Assembly references and imported namespaces for internal implementation""  
xmlns=""http://schemas.microsoft.com/netfx/2009/xaml/activities""  
xmlns:mc=""http://schemas.openxmlformats.org/markup-compatibility/2006""  x
mlns:mva=""clr-namespace:Microsoft.VisualBasic.Activities;assembly=System.Activities""  
xmlns:scg=""clr-namespace:System.Collections.Generic;assembly=mscorlib""  
[...]

Issue #3, opened 2016-09-14
DynamicActivity


ewinnington:
Will DynamicActivity be ported to CoreWF? Or is the functionality covered in some other way?

This would allow "To create an activity at runtime using imperative code" as described in the upper part of https://msdn.microsoft.com/en-us/library/ee358722(v=vs.110).aspx

dmetzgar:
DynamicActivity takes a heavy dependency on System.ComponentModel, most of which is not ported to Core. There are some parts of componentmodel, but not enough to cover WF scenarios. It would have to be evaluated to see if regular reflection could be used instead.

dmetzgar:
Looks like .NET Standard 2.0 is planning to have PropertyDescriptor and TypeDescriptor. Those are key for porting dynamicactivity and automatic cachemetadata.

ewinnington:
With 2.0 arriving soon, is there something external contributors could do to help port coreWF?

dmetzgar:
I updated the readme with a bit more details. Would you mind having a look?

ewinnington:
Thanks for the update. So if I understand correctly, we can fork your repo and start bringing in the DynamicActivity and the ActivityXamlServices from the MIT licensed [ReferenceSource](https://github.com/Microsoft/referencesource), get that to compile with NetStandard 2.0, add [PortableXAML](https://github.com/cwensley/Portable.Xaml) for the parsing and we are off to the races?

Thank you again for that update.

dmetzgar:
Yes, that's right! I'll be helping out here and there but all of this work is definitely up for grabs.

ewinnington:
I've started going through adding DynamicActivity and re-enabling attributes such as [Ambient] through pulling in PortableXAML. 

Was there any documentation or porting information in the move to net core? 
A list of the remappping of the namespaces perhaps? Microsoft.Activities to CoreWf. 

I'll diff with the reference source to try and understand the changes.

ewinnington:
I've brought DynamicActivity without any XAML namespace into coreWF. It compiles, but does not work for the simple example of ["To create an activity at runtime using imperative code" ](https://docs.microsoft.com/en-us/dotnet/framework/windows-workflow-foundation/creating-an-activity-at-runtime-with-dynamicactivity). 

I'm getting this exception: 

Unhandled Exception: CoreWf.InvalidWorkflowException: The following errors were encountered while processing the workflow tree:
[...]

PR #25, opened 2018-08-01
Xaml

dmetzgar:
This is basically restoring all the code for WF around XAML integration and transactions. Portable.Xaml is used in place of System.Xaml. 

System.Transactions is included with .NET Core 2.0, but it is not part of the .NET Standard. Therefore, the transactions code is copied to CoreWf.Transactions. 

Some features like debugging support and dynamic update are left out of compilation. Although I marked these as targeting .NET 4.5, I've not tested with targeting that framework. 

This code also uses Microsoft.Extensions.Localization for string resource handling.


dmetzgar:
Fixes #16 
Fixes #14 
Fixes #6 
Fixes #4 
Fixes #3

Diaskhan:
@dmetzgar hie, what is main prioprity of  project? 
It possible to make bpm notation for ure engine ? 

What help is needed ?

Diaskhan:
Actually i need bpm engine for .net! 
Its possible to make some layer that executes bpm models over ure engine ?

dmetzgar:
@Diaskhan, I'm not familiar enough with BPM to know if there is parity with the activities available in WF and those in the BPEL spec. WF was built around XAML but it's entirely possible to make it work with a different format. My main goal for this project is to get parity with WF in the .NET Framework for most features. After that, BPM is a possibility but I'm not sure of the demand for it. Can you describe your scenario that you need BPM support?

Diaskhan:
Something similar to https://camunda.com/
Is old Activiti engine!

But camunda is bpm System. Not ECM!
I want something similar to camunda! But I hate Java))

There have designer, that could write bpm files! Its open source!
And we could implement self designer of bpm files for U're engine !
https://bpmn.io/

By workflow conception we could create,crm,bpm,ecm systems! 

I have maded analitics for USSR market of ecm systems. Many of them uses WWF! And there engine work with bpm like notation!

Main components is , humantask,scripttask,gateways,suprocceses and any custom nodes that realized custom logics of the system!

I'am advice to U to download some bpm system. And analyze, it. BPEL is very old conception!

Bpm is everywhere!
I have played with this  ECM,CRM,BPM system! 
http://www.elma-bpm.com/

bpmonline is my job ! We are customizing it as Integration Partner! Is CRM system!
https://www.bpmonline.com/

Diaskhan:
Actually right now i'm working to make DAL LAyer for enterprise system,
After that Im gonna diving in workflow engine!
image

ewinnington:
@dmetzgar By modifying the Activity Implementation, I was able to load and parse in dynamic activities two simple XAML Workflows (Empty and WriteLine) 

See my commit: [Parsed successfully XAML -> DynamicActivity](https://github.com/ewinnington/corewf/commit/200674d84cb07e05e0ceac1033aad8c7a543b6ed)

- Changed the visibility of Activity Implementation from protected to public allowed Portable.Xaml to reach the property and fill the Implementation body.

This allows:
```
 <Activity x:Class=""WFTemplate""
            xmlns=""http://schemas.microsoft.com/netfx/2009/xaml/activities""
            xmlns:x=""http://schemas.microsoft.com/winfx/2006/xaml"">
<Sequence>
            <WriteLine Text=""HelloWorld"" />
</Sequence>
</Activity>

```
To be successfully parsed and functional. Writes out "HelloWorld".

I've added my test code to my[ XAML branch. ](https://github.com/ewinnington/corewf/tree/xaml)

The next step would be lighting up compilation to allow the [MyInput] InputArguments to be applied. Currently, it just writes out [MyInput] in the WriteLine, so I guess the variable parsing and compilation is not active.

ewinnington:
I've also checked my previous Imperative code workflows, they also work with this XAML branch. This allows: https://docs.microsoft.com/en-us/dotnet/framework/windows-workflow-foundation/creating-an-activity-at-runtime-with-dynamicactivity

So looking good so far.

Also added the[ Imperative version of the Transacted test scope](https://github.com/ewinnington/corewf/commit/6389b8a4c818562a2a94d34915511229925a6370) from the [MS Documentation Workflow Transactions](https://docs.microsoft.com/en-us/dotnet/framework/windows-workflow-foundation/workflow-transactions) and it works out of the box, so it seems at least the basics of transactions is working too.  Looking good for #14

ewinnington:
The previous branch had 4 failing tests (*), the XAML branch has 9 failing tests, the 4 previous and 5 new ones which are:
TestCases.Runtime.WorkflowInstanceTest.WorflowInstanceResumeBookmarkAsyncTests.TestPersistDuringResumeBookmark
TestCases.Activities.Rethrow.PersistAfterCatchBeforeRethrow
TestCases.Activities.Flowchart.Execution.UnloadFlowchartWhileExecutingFlowStep
TestCases.Activities.Flowchart.Execution.UnloadFlowchartWhileExecutingFlowSwitchExpression
TestCases.Activities.Flowchart.Execution.UnloadFlowchartWhileExecutingFlowConditionalCondition

I'm tracking this down, but it all seems to be related to the JSON Instance store, failing deserialization somewhere. Either this results in a:
Result Message:	System.IO.FileNotFoundException : Could not load the specified file.
Or a: 
Newtonsoft.Json.JsonSerializationException : Could not convert string 'Block' to dictionary key type 'CoreWf.Bookmark'. Create a TypeConverter to convert from the string to the key type object. Path '['{urn:schemas-microsoft-com:CoreWf/4.0/properties}Workflow'].Value.bookmarkMgr.bookmarks.Block', line 14, position 18.
---- Newtonsoft.Json.JsonSerializationException : Error converting value "Block" to type 'CoreWf.Bookmark'. Path '['{urn:schemas-microsoft-com:CoreWf/4.0/properties}Workflow'].Value.bookmarkMgr.bookmarks.Block', line 14, position 18.
-------- System.ArgumentException : Could not cast or convert from System.String to CoreWf.Bookmark.

I'll try digging a little more to see if I can fix those.

--------------------------
(*) 2 of those are due to my DateTimeFormat being different from the one that is expected:
Test.TestCases.Activities.Expressions.Add.AddTwoCompatibleDifferentTypes
Message: System.Exception : Validation errors:
Step(s) is not found UserTrace: 2/12/2009 5:58:59 PM (actual 0/req 1/opt 0)
Unexpected step(s) found UserTrace: 12.02.2009 17:58:59 (actual 1/req 0/opt 0)

Test.TestCases.Activities.Expressions.Subtract.SubtractTwoCompatibleDifferentTypes
System.Exception : Validation errors:
Step(s) is not found UserTrace: 2/11/2009 6:01:01 AM (actual 0/req 1/opt 0)
Unexpected step(s) found UserTrace: 11.02.2009 06:01:01 (actual 1/req 0/opt 0)

And the other two complain about: 
Message: System.Resources.MissingManifestResourceException : Could not find any resources appropriate for the specified culture or the neutral culture.  Make sure "CoreWf.Strings.SR.resources" was correctly embedded or linked into assembly "CoreWf" at compile time, or that all the satellite assemblies required are loadable and fully signed.

galvesribeiro:
@Diaskhan ,

Think of WF as the primitives to any workflow-based tech. Multiple companies today offer BPMN tools based on the _legacy_ WF for .Net Framework in Sharepoint. If you want BPMN modeling using WF as a runtime for it, it is perfectly fine. However, I don't think it is something that should be _inside_ WF framework.

If I may suggest, please open an issue with your ideas and requests for BPMN and leave this PR for the reviewers so we don't pollute it with non-related discussion.

The purpose of this PR is bring the XAML support back to the WF Core and not introduce something new.

Don't take me bad, I always thought BPMN a nice candidate to be used with WF. I just think that this PR is not the right forum for discussion related to it.

Thanks!

ewinnington:
By adding [parameterless constructors for the JSON deserialization](https://github.com/ewinnington/corewf/commit/d3487000d7e08e7a65c0398e23daca4af6bcc26d) to the following
src/CoreWf/ActivityInstance.cs
src/CoreWf/Bookmark.cs
src/CoreWf/Hosting/BookmarkInfo.cs
src/CoreWf/Runtime/ActivityExecutor.cs
src/CoreWf/Runtime/BookmarkCallbackWrapper.cs
src/CoreWf/Runtime/CallbackWrapper.cs
src/CoreWf/Runtime/LocationEnvironment.cs

I get the test cases further to a validation error at the end of the run. There's something that is broken with the TestBlockingActivity (And/Or Bookmarks). 
Message: System.Exception : Validation errors:
Unexpected step(s) found ActivityTrace: Flowchart27, Executing (actual 2/req 1/opt 0)
Unexpected step(s) found ActivityTrace: Block, Executing (actual 2/req 1/opt 0)

@dmetzgar I hope this helps in finding why those tests fail.

dmetzgar:
@ewinnington thanks for investigating this! Could you please make a PR against the xaml branch? This will help a lot.

ewinnington:
@dmetzgar PR #26 is up. Sorry, I could have given a better name.

dmetzgar:
It looks like most of the test failures left are caused by serialization through JSON. With XAML back in the picture, the custom serialization done through the NetDataContractSerializer is needed. It can probably be mimicked with JSON through custom serializers. I don't believe NetDCS is available in .NET Standard so I'll need to do some research on how to get the serialization right.

@ewinnington, can you turn your imperative and xaml tests into xunit test cases?

ewinnington:
Yes, I’ll have a look to delete / clean up, my mess and make them into proper tests. 

By the way, I fixed Test.TestCases.Activities.Expressions.Add.AddTwoCompatibleDifferentTypes and Substract test cases on my branch, it was the regional settings of the toString() that was causing the issue, since it was trying to compare to a string containing a date in US format.

dmetzgar:
Cool. Thanks for finding that.

dmetzgar:
Quick update: NetDCS is not available in .NET Core (discussion on that here: https://github.com/dotnet/corefx/issues/6564). I tried regular DCS but that's not cutting it and neither is JSON.NET's serializer. In that linked issue, Rocky Lhotka talks about creating his own serializer for CSLA. 

It looks like I'll have to use BinarySerializer. There are some weird tricks that WF was using that are best handled by NetDCS. The biggest concern I have for creating switching serializers is the incompatibility with existing serialized workflow instances. In order to migrate to CoreWF, a user would need to target .NET Framework first and CoreWF would have to fallback to using NetDCS for deserialization if the custom serializer didn't work. Then the user would need to make sure to load and re-persist all instances in their data store. Once that's done, then they can retarget their app for Core.

Another issue is that many types are not marked `[Serializable]` and BinarySerializer might not work. I'll try it out this weekend and see if I can get it to work for the tests. Here's the issue I'm talking about: https://github.com/dotnet/corefx/issues/19119

PR #26, opened 2018-08-03
Xaml

ewinnington:
Allowed the DynamicActivity to be parsed from Xaml
Added parameterldss constructors for the Json serialization

— added a couple of test console apps that could be removed / consolidated


dmetzgar (approved):
Great! Thanks for adding the tests as well!

PR #47, opened 2019-02-14
Compile Vb expressions using Roslyn

lbargaoanu:
The only interesting code is in [HostedCompiler.cs](https://github.com/dmetzgar/corewf/blob/493ba2ce95992e06550aaea09ae515250141d6ca/src/CoreWf/Microsoft/VisualBasic/Activities/HostedCompiler.cs). I'm guessing the relevant tests are closed source.

### DIFF
--- a/src/CoreWf/CoreWf.csproj
+++ b/src/CoreWf/CoreWf.csproj
@@ -31,6 +31,7 @@
     </PackageReference>
     <PackageReference Include="Microsoft.Extensions.Localization" Version="2.1.1" />
     <PackageReference Include="Portable.Xaml" Version="0.23.0" />
+    <PackageReference Include="System.CodeDom" Version="4.5.0" />
     <PackageReference Include="System.ComponentModel.EventBasedAsync" Version="4.3.0" />
     <PackageReference Include="System.Reflection.TypeExtensions" Version="4.5.0" />
     <PackageReference Include="System.Runtime.Serialization.Primitives" Version="4.3.0" />
@@ -62,4 +63,13 @@
       <BuildOutputInPackage Include="@(_ResolvedProjectReferencePaths)" />
     </ItemGroup>
    </Target>
+  <ItemGroup>
+    <Compile Include="Debugger\SourceLocation.cs" />
+    <Compile Include="Debugger\SourceLocationProvider.cs" />
+    <Compile Include="Debugger\Symbol\ActivitySymbol.cs" />
+    <Compile Include="Debugger\Symbol\DebugSymbol.cs" />
+    <Compile Include="Debugger\Symbol\SymbolHelper.cs" />
+    <Compile Include="Debugger\Symbol\WorkflowSymbol.cs" />
+    <Compile Include="ExpressionParser\SourceExpressionException.cs" />
+  </ItemGroup>
 </Project>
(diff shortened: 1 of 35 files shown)


ewinnington:
First of all, thank you!

ewinnington:
You are a wizard! This is amazing 💯 !

HelloWorldConsole : CodeActivity<bool> doesn't seem to work correctly to get the input argument, but maybe it's wrong code from me in the test. 

edit: Yeah, HelloWorldConsole was my mistake. It's all good. I need to update that piece of code because I got the if condition inverted.

ewinnington:
After a check, additional testing like [MyInput + 10] - Direct math , [MyInput + DateTime.Now.Month] - Static object with property and [myInput+CINT(DateTime.Now().GetTypeCode())] - conversions with function calls all work in the XAML. 

@dmetzgar If you are ok with the code submitted, I think this should be merged. We might need to see if we can find additional tests / checks to ensure the code works.

src/CoreWf/CoreWf.csproj
```diff
@@ -25,15 +25,21 @@
     <AssemblyOriginatorKeyFile>..\..\CoreWF.snk</AssemblyOriginatorKeyFile>
   </PropertyGroup>
   <ItemGroup>
-    <PackageReference Include="Microsoft.CodeAnalysis.FxCopAnalyzers" Version="2.6.1">
+    <PackageReference Include="Microsoft.CodeAnalysis.FxCopAnalyzers" Version="2.6.2">
       <PrivateAssets>all</PrivateAssets>
       <IncludeAssets>runtime; build; native; contentfiles; analyzers</IncludeAssets>
     </PackageReference>
+    <PackageReference Include="Microsoft.CodeAnalysis.VisualBasic.Scripting" Version="2.8.0-dev-62823-01" />
```
ewinnington, line 32: This is from a MyGet source. Also I get a compatibility message, is there a way to avoid that? Is the Roslyn package available at that version on nuget somehow?

lbargaoanu: Not that I'm aware of.

ewinnington: @dmetzgar I'd like your input on adding that as a source before we merge this in. Is it OK to add this dependency? It is an MS package after all, just not (yet?) at that version on nuget.

@lbargaoanu Could this work with 
nuget install Microsoft.Net.Compilers   # Install C# and VB compilers
nuget install Microsoft.CodeAnalysis    # Install Language APIs and Services
from the https://github.com/dotnet/roslyn repo or do we need the custom one from myget?

ewinnington: I tried adding all the nugets from Microsoft.CodeAnalysis.*, but couldn't find the Microsoft.CodeAnalysis.VisualBasic.Scripting - VisualBasicScript class. So it seems we require that MyGet feed for CoreWF.

dmetzgar:
@lbargaoanu this is great! By relevant tests being closed source, are you referring to the original tests we had in the .NET Framework? I just haven't finished porting them to run in this test framework.

lbargaoanu:
Yes, that's what I meant. If we can get them, I can try converting them. If they're in the repo somewhere, I didn't find them.

Issue #49, opened 2019-03-20
Microsoft.CodeAnalysis.VisualBasic.Scripting cannot be restored


ThomasZitzler:
I tried to compile CoreWf, but nuget cannot restore the "Microsoft.CodeAnalysis.VisualBasic.Scripting" package. 

Where should that come from?

ewinnington:
It requires the Roslyn feed from MyGet (See https://github.com/dotnet/roslyn ) 
https://dotnet.myget.org/F/roslyn/api/v3/index.json

Sorry about that, it's the new VB / C# compilation that requires it for the Dynamic XAML.

ThomasZitzler:
Thanks, I added a nuget.config with that package source. now it's compiling, but I get the following warnings:
```
CoreWf.csproj : warning NU1603: CoreWf depends on Microsoft.CodeAnalysis.VisualBasic.Scripting (>= 2.8.0-dev-62823-01) but Microsoft.CodeAnalysis.VisualBasic.Scripting 2.8.0-dev-62823-01 was not found. An approximate best match of Microsoft.CodeAnalysis.VisualBasic.Scripting 2.8.1-beta6-62904-04 was resolved.
CoreWf.csproj : warning NU1701: Package 'Microsoft.CodeAnalysis.VisualBasic.Scripting 2.8.1-beta6-62904-04' was restored using '.NETFramework,Version=v4.6.1' instead of the project target framework '.NETStandard,Version=v2.0'. This package may not be fully compatible with your project.

```

So does it mean it now falls back to the full framework?

ewinnington:
I’ve been using it on core, haven’t found something that throws a not supported or anything yet.

ThomasZitzler:
See here: https://github.com/dotnet/roslyn/issues/13523

It seems there will be no official version...

ThomasZitzler:
@ewinnington Did you try to compile/run it on linux? The whole porting of WF only makes sense when it's not platform/.net framework dependent. 

My main reason for using corewf would be to run our program in a linux docker container.

ewinnington:
No, I was testing on windows using core. It should work everywhere. We might have to create that nuget for .net standard ourselves and check if there is any incompatibility. 

There is a Workflow runner working on core on https://github.com/ewinnington/wfRunnerCore that I’m using to test workflows. I’ll clone it to a Linux this weekend when I have a minute to myself.

Also thanks for the pull request. I’ll test it and merge it this weekend.

[...]

PR #50, opened 2019-03-21
Create nuget.config

ThomasZitzler:
The nuget.config allows to get the missing VisualBasic.Scripting package, see #49


lbargaoanu:
Thanks, but I've already added a minimal nuget.config.

PR #52, opened 2019-03-22
vb: support complex parameters with typename

ThomasZitzler:


src/XamlTestCases/HelloWorldConsole2.cs
```diff
@@ -0,0 +1,34 @@
+﻿using System;
+using System.Activities;
+
+namespace XamlTestCases
+{
+    public class HelloWorld2Input
```
lbargaoanu: I'm sure we can find better names here :)

src/XamlTestCases/XamlTests.cs
```diff
@@ -174,5 +174,29 @@ public void XamlWorkflowWithInputsOutputs()
...
+
+        [Fact]
+        public void XamlWorkflowWithInputObject()
+        {
+            var xamlString = @"
+                    <Activity x:Class=""WFTemplate""
+                              xmlns=""http://schemas.microsoft.com/netfx/2009/xaml/activities""
+                              xmlns:s=""clr-namespace:System;assembly=mscorlib""
+                              xmlns:s1=""clr-namespace:System;assembly=System""
+                              xmlns:sa=""clr-namespace:CoreWf;assembly=CoreWf""
```
lbargaoanu: Does it work without this line? Most things were renamed to System.Activities.

lbargaoanu:
Looks good, but a few minor things.

src/XamlTestCases/XamlTests.cs
ThomasZitzler: You are right, I started with it before you renamed and did copy + paste of the old one to get started. Works with "System.Activities"

lbargaoanu (approved):
Thanks!

Issue #55, opened 2019-05-29
WorkflowApplication.ResumeBookmark starts workflow from beginning


zehavibarak:
Resuming a bookmark should not call Activity.Execute at all. Instead, it should only call the callback.

zehavibarak:
I'm using a guid bookmark for context.CreateBookmark(...). The resume repeats this call, overriding the old bookmark with the new one.

lbargaoanu:
A PR is welcome.

zehavibarak:
The ResumeBookmark below calls the activity Execute, this is wrong
```c#
            WorkflowApplication wfApp = CreateWorkflowApplication();
            var workflowInstanceId = wfApp.Id;
            wfApp.Run();
            string bookmarkName = workflowInstanceId.ToString();
          _unloadedEvent.WaitOne();
/* create */                    
             wfApp = CreateWorkflowApplication();
             wfApp.Load(workflowInstanceId);
             var result = wfApp.ResumeBookmark(bookmarkName, null); // <- this hits the BookmarkActivity while **IT SHOULD'NT**

        public static WorkflowApplication CreateWorkflowApplication()
        {
            Activity wf = CreateWorkflow();
            WorkflowApplication result = new WorkflowApplication(wf)
            {
                Unloaded = e =>
                {
                    _unloadedEvent.Set();
                },
                PersistableIdle = e => PersistableIdleAction.None
            };

            return result;
        }
        
        public static Activity CreateWorkflow()
        {
            Sequence workflow = new Sequence();
            workflow.Activities.Add(
                new WriteLine
                {
                    Text = "Before Bookmark"
                });
            workflow.Activities.Add(new BookmarkActivity());
            workflow.Activities.Add(
                new WriteLine
                {
                    Text = "After Bookmark"
                });

            return workflow;
        }
    

public class BookmarkActivity : NativeActivity
    {
        protected override void Execute(NativeActivityContext context)
        {
            Guid wfInstanceId = context.WorkflowInstanceId;
            string bookmarkName = wfInstanceId.ToString();

            Console.WriteLine("BookmarkActivity.Execute - creating bookmark with name {0}", bookmarkName);
            context.CreateBookmark(bookmarkName, new BookmarkCallback(BookmarkCallback), BookmarkOptions.MultipleResume);
        }

        protected override bool CanInduceIdle => true;

        private void BookmarkCallback(NativeActivityContext context, Bookmark bookmark, object bookmarkData)
        {
[...]

Issue #31, opened 2018-09-16
TestPersistDuringResumeBookmark test fails


dmetzgar:
Test is [here](https://github.com/dmetzgar/corewf/blob/678af1626bd825cfff5f0d6d0d72196fe62853df/test/TestCases.Runtime/WorflowInstanceResumeBookmarkAsyncTests.cs#L302)

Issue #29, opened 2018-09-16
Tests using bookmarks don't serialize correctly


dmetzgar:
Tests like [UnloadFlowchartWhileExecutingFlowConditionalCondition](https://github.com/dmetzgar/corewf/blob/678af1626bd825cfff5f0d6d0d72196fe62853df/test/TestCases.Activities/Flowchart/Execution.cs#L544) use a TestBlockingActivity that doesn't serialize correctly in the JSON serializer:
```csharp
                Body = new TestBlockingActivity("Block");
```

zehavibarak:
Add to DurableInstancing InstanceValue:
```c#
    [KnownType(typeof(ActivityExecutor))]
    [KnownType(typeof(Variable<int>.VariableLocation))]
    [KnownType(typeof(Variable<string>.VariableLocation))]
```
in ActivityInstance, add IsReference = true
```c#
    [DataContract(Name = XD.ActivityInstance.Name, Namespace = XD.Runtime.Namespace, IsReference = true)]
```

PR #57, opened 2019-06-14
Allows resuming Wf at correct point after persistence

ewinnington:
Added InitializeCore(null, null) back into the codepath in WorkflowInstance

Thanks to @zehavibarak for the information on the missing InitializeCore.
This was part of the commented #if NET45 in the reload scenarios. (See comment // used for Load scenarios where you are rehydrating a WorkflowInstance).

Added a BookmarkConverter (new class, new file), so that we have less issues with the jsonstore.

This allows the following previously skipped tests to now pass:
UnloadFlowchartWhileExecutingFlowConditionalCondition
UnloadFlowchartWhileExecutingFlowSwitchExpression
UnloadFlowchartWhileExecutingFlowStep

The two tests which were flaky now seem to pass:
TestPersistDuringResumeBookmark
TestAbortLoad

These two tests are still timing dependent and always pass if run singly. They currently run on my RunAll.

### DIFF
--- a/src/CoreWf/Bookmark.cs
+++ b/src/CoreWf/Bookmark.cs
@@ -10,9 +10,11 @@ namespace System.Activities
     using System.Runtime.Serialization;
     using System.Globalization;
     using System.Activities.Internals;
+    using System.ComponentModel;
 
     [DataContract]
     [Fx.Tag.XamlVisible(false)]
+    [TypeConverter(typeof(BookmarkConverter))]
     public class Bookmark : IEquatable<Bookmark>
     {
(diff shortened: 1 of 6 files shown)


ewinnington:
Should fix issues #55 , #31 , #29

ewinnington:
@lbargaoanu Any suggestions before merging? Are you OK with the addition of the BookmarkConverter?

lbargaoanu:
I'll take a look today if I have the time.

lbargaoanu (approved):

PR #62, opened 2019-09-12
fix NullReferenceException of DelayTimer after load from instance store

LunicLynx:
## Fix `NullReferenceException` after load from `InstanceStore`

After load from instance store with an expired timer, the if condition `timerData.ExpirationTime <= DateTime.UtcNow` can be true. 
This will leave `timerData.DelayTimer` uninitialized (`null`). `TimerTable` then tries to remove the timer in `TimerTable.RemoveTimer` resulting in a `NullReferenceException` here
https://github.com/UiPath/corewf/blob/66da71b919c6597cbcc61578d10ae97503476c5e/src/CoreWf/Statements/TimerTable.cs#L76

Note: In combination with `LoadRunnableInstance()`.

### DIFF
--- a/src/CoreWf/Bookmark.cs
+++ b/src/CoreWf/Bookmark.cs
@@ -10,9 +10,11 @@ namespace System.Activities
     using System.Runtime.Serialization;
     using System.Globalization;
     using System.Activities.Internals;
+    using System.ComponentModel;
 
     [DataContract]
     [Fx.Tag.XamlVisible(false)]
+    [TypeConverter(typeof(BookmarkConverter))]
     public class Bookmark : IEquatable<Bookmark>
     {
         private static readonly Bookmark asyncOperationCompletionBookmark = new Bookmark(-1);
--- a/src/CoreWf/BookmarkConverter.cs
+++ b/src/CoreWf/BookmarkConverter.cs
@@ -1,0 +1,34 @@
+﻿// This file is part of Core WF which is licensed under the MIT license.
+// See LICENSE file in the project root for full license information.
+
+namespace System.Activities
+{
+    using System;
+    using System.Globalization;
+    using System.ComponentModel;
+
+    public class BookmarkConverter : TypeConverter
+    {
+        public override bool CanConvertFrom(ITypeDescriptorContext context, Type sourceType)
+        {
+            return sourceType == typeof(string) || sourceType == typeof(long); 
+        }
+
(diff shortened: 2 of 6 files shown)


LunicLynx:
@lbargaoanu i had another look into this. If you can give me some guidance i might be able to create the tests you suggested.
Currently there is no access to internals as far as i can see. I think i would need access to at least `IPersistencePipelineModule` to invoke the initialization. Or how do you go about these scenarios?

lbargaoanu:
There are more than a few tests that use `TestWorkflowRuntime`, Persist and Resume. Try doing the same.

LunicLynx:
Yes but timers are different. The `TryLoadRunnableWorkflowCommand` isn't yet implemented in the `FileInstanceStore` either. 

But I think I almost done it. Currently i get this error: 

```
Unexpected step(s) found WorkflowInstanceTrace: Unloaded (actual 1/req 0/opt 0)
```

My test looks like this:

```csharp

TestSequence seq = new TestSequence
            {
                Activities =
                {
                    new TestWriteLine("Before") {Message = "Before"},
                    new TestDelay()
                    {
                        Duration = TimeSpan.FromSeconds(1)
                    },
                    new TestWriteLine("After") {Message = "After"}
                }
            };

 using (TestWorkflowRuntime runtime =
                TestRuntime.CreateTestWorkflowRuntime(seq, null, jsonStore, PersistableIdleAction.Unload))
            {
                runtime.ExecuteWorkflow();

                runtime.WaitForUnloaded();

                Thread.Sleep(TimeSpan.FromSeconds(2));
               
                runtime.LoadRunnableWorkflow();

                runtime.ResumeWorkflow();

                runtime.WaitForCompletion();


            }
```

LunicLynx:
What do i need to do so that the expected trace contains an unloaded step?

lbargaoanu:
No idea :) But the existing tests do work...

LunicLynx:
Who need I ask for advice then?

lbargaoanu:
You just need to figure it out.

LunicLynx:
Aren’t you interested in a solution? 
I could compile it for my use case and just move on you know, but that seems not in the sprit of Open source.
Your behavior seems almost a bit hostile. I just wanted to help.

lbargaoanu:
There is nobody to help. If I start looking into this, it's easier to do it myself. But I'm not going to do that. So what would help is for you to understand how existing tests work and write one that applies to your case. I can see that you're looking for a shortcut, but there isn't any.

PR #64, opened 2019-10-11
Target .Net Core 3.0 desktop

### LINKED ISSUE #54: Replace Portable.Xaml with System.Xaml
We need https://github.com/dotnet/wpf/issues/46.

lbargaoanu:
Fixes #54.


ewinnington:
Targeting netcoreapp3.0 with wpf does mean we lose out of x-platform ability, at least until we can use system.xaml as a separate nuget component.

lbargaoanu:
Yes, but I've no better idea. This is a clean way to bring System.Xaml into the mix. In theory System.Xaml shouldn't depend on anything Windows specific, but who knows :)

Issue #19, opened 2017-07-19
Port remaining WF tests


dmetzgar:
A lot of tests were left out because the features weren't implemented. Those tests can be ported and then disabled. As features are added, we can enable their tests.

ewinnington:
This will be very helpful to check my progress and the validity of the modifications.

dmetzgar:
There are no more tests for to port for the current capabilities of CoreWF. A lot of the other tests were on WF services, request/reply, etc., which use WCF underneath.

PR #75, opened 2020-02-13
Add round trip XAML testing

dmetzgar:
This PR goes toward issue #19. Part of the original framework tests was to take a given activity and populate its properties with randomized values then serialize to XAML, deserialize, and compare the objects before and after. There is some special casing, such as for arguments. Not all cases are handled yet and some activities use dynamic implementations that currently are not serializable with the test code.

This code will also be helpful for testing that the System.Xaml library used on non-Windows platforms works correctly.


src/Test/TestCases.Xaml/TestCases.Xaml.csproj
```diff
@@ -0,0 +1,30 @@
...
+
+  <PropertyGroup>
+    <TargetFrameworks>netcoreapp3.0</TargetFrameworks>
+
+    <IsPackable>false</IsPackable>
+
+    <RunAnalyzersDuringBuild>false</RunAnalyzersDuringBuild>
+  </PropertyGroup>
+
+  <PropertyGroup Condition="'$(Configuration)|$(Platform)'=='Release|AnyCPU'">
```
lbargaoanu: Just remove these two I guess.

src/Test/TestCases.Xaml/TestCases.Xaml.csproj
```diff
@@ -0,0 +1,30 @@
+<Project Sdk="Microsoft.NET.Sdk">
```
lbargaoanu, line 1: We need to do smth about the XamlTest projects. Perhaps rename the existing one to TestCases.Workflows?

src/CoreWf/Expressions/AndAlso.cs
```diff
@@ -10,7 +10,9 @@ namespace System.Activities.Expressions
 using System.Activities.DynamicUpdate;
 
     //[SuppressMessage(FxCop.Category.Naming, FxCop.Rule.IdentifiersShouldNotMatchKeywords, Justification = "Optimizing for XAML naming. VB imperative users will [] qualify (e.g. New [AndAlso])")]
+#pragma warning disable CA1716 // Identifiers should not match keywords
```
lbargaoanu: Wouldn't [this](https://docs.microsoft.com/en-us/dotnet/api/system.diagnostics.codeanalysis.suppressmessageattribute?view=netframework-4.8) work? Or any other less painful syntax :)

src/Test/TestCases.Xaml/TestCases.Xaml.csproj
```diff
@@ -0,0 +1,30 @@
...
+
+  <PropertyGroup>
+    <TargetFrameworks>netcoreapp3.0</TargetFrameworks>
```
lbargaoanu: This should also target netcoreapp2.1 to check the portable System.Xaml.

src/CoreWf/Expressions/AndAlso.cs
lbargaoanu: Ok, that was the initial code :) I think, just remove the pragma for now. Perhaps we should enable the attributes at some point.

src/Test/TestCases.Xaml/TestCases.Xaml.csproj
dmetzgar: Done

src/Test/TestCases.Xaml/TestCases.Xaml.csproj
dmetzgar: Done

src/CoreWf/Expressions/AndAlso.cs
dmetzgar: Removed

src/Test/TestCases.Xaml/TestCases.Xaml.csproj
dmetzgar: Let me handle that in a separate PR. Created issue #76

PR #81, opened 2020-02-20
Uncomment xaml tests

lbargaoanu:


lbargaoanu:
There are still problems with the WriteLine tests, but I don't have a fix.

Issue #86, opened 2020-03-12
VB Expressions fail with netcore3


dmetzgar:
A test that use VB expressions in a XAML that is loaded with ActivityXamlServices.Load fails. The test is [here](https://github.com/UiPath/corewf/blob/b38ca67bfb0ae1206bbfe65deb629a3fbfa63a0d/src/Test/TestCases.Workflows/WF4Samples/Expressions.cs#L28).

The failure is "error BC35000: Requested operation is not available because the runtime library function 'Microsoft.VisualBasic.CompilerServices.ProjectData.SetProjectError' is not defined."

Full stack is below:
```
    System.Activities.InvalidWorkflowException : The following errors were encountered while processing the workflow tree:
    'DynamicActivity': The private implementation of activity '1: DynamicActivity' has the following validation error:   Compiler error(s) encountered processing expression "New SalaryStats()".Microsoft.CodeAnalysis.Scripting.CompilationErrorException: (1) : error BC35000: Requested operation is not available because the runtime library function 'Microsoft.VisualBasic.CompilerServices.ProjectData.SetProjectError' is not defined.
       at Microsoft.CodeAnalysis.Scripting.ScriptBuilder.ThrowIfAnyCompilationErrors(DiagnosticBag diagnostics, DiagnosticFormatter formatter)
       at Microsoft.CodeAnalysis.Scripting.ScriptBuilder.CreateExecutor[T](ScriptCompiler compiler, Compilation compilation, Boolean emitDebugInformation, CancellationToken cancellationToken)
       at Microsoft.CodeAnalysis.Scripting.Script`1.GetExecutor(CancellationToken cancellationToken)
       at Microsoft.CodeAnalysis.Scripting.Script`1.CommonGetExecutor(CancellationToken cancellationToken)
       at Microsoft.CodeAnalysis.Scripting.Script`1.TryGetPrecedingExecutors(Script lastExecutedScriptInChainOpt, CancellationToken cancellationToken)
       at Microsoft.CodeAnalysis.Scripting.Script`1.GetPrecedingExecutors(CancellationToken cancellationToken)
       at Microsoft.CodeAnalysis.Scripting.Script`1.RunAsync(Object globals, Func`2 catchException, CancellationToken cancellationToken)
[...]

PR #87, opened 2020-03-12
Adding WF 4.0 Samples

dmetzgar:
After examining the remaining WF tests from the framework (#19), there are only a handful of tests that can be ported and they're not covering new areas. Since we need more coverage of XAML with VB expressions, this PR adds tests from the [WF Samples](https://social.technet.microsoft.com/wiki/contents/articles/12326.windows-workflow-wf-4-x-samples.aspx).


lbargaoanu:
Merging, as this is getting too far ahead of master. Fixes #86.

PR #89, opened 2020-03-15
WIP: Compile with roslyn through codedom provider

### LINKED ISSUE #48: CSharpValue and CSharpReference are not fully functional yet
When working on Xaml workflows generated by visual studio with CSharpValue and CSharpReference elements, I get a Workflow not compiled error even after setting compilation to true in the settings. 

I’ll try and resolve this down to a simple test case, the WF I built was too ambitious. 

Has anyone else managed to use a Visual Studio defined Xaml Workflow? 

Also with the assembly name change, you have to change the Xaml declaration at the beginning of the Xaml to make it work. 

This is related to my testing this in my [wfRunnerCore repo](
https://github.com/ewinnington/wfRunnerCore).

Related to #7

ThomasZitzler:
This is an experiment of mine to get CSharpValue and CSharpReference up and running to fix #48 

It still uses the code dom provider to generate the code, but then passes that code to the roslyn compiler for compilation.

Unfortunately I was not able to find a good combination of the packages together with the VB scripting prerelease version, so I also converted the VB value + reference so that they always require compilation.

For my project that would be good enough, we anyway only use it that way.

Tests seem to pass that far, unfortunately cannot try it on linux as System.Activities is not compilable there (becuase of the desktop sdk).

Please let me know if you have some ideas/proposals.

### DIFF
--- a/src/CoreWf/Microsoft/VisualBasic/Activities/HostedCompiler.cs
+++ b/src/CoreWf/Microsoft/VisualBasic/Activities/HostedCompiler.cs
@@ -2,7 +2,7 @@
 using Microsoft.CodeAnalysis.Scripting;
 using Microsoft.CodeAnalysis.Scripting.Hosting;
 using Microsoft.CodeAnalysis.VisualBasic;
-using Microsoft.CodeAnalysis.VisualBasic.Scripting;
+//using Microsoft.CodeAnalysis.VisualBasic.Scripting;
 using Microsoft.CodeAnalysis.VisualBasic.Syntax;
 using ReflectionMagic;
 using System;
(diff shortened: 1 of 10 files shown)


lbargaoanu:
It's not acceptable to _require_ compilation for the VB expressions. I mean, if you want to work that way, fine, but the main scenario must still work. If you cannot do it with the preview Roslyn package, you should rewrite the code so it's not needed.

lbargaoanu:
I guess the problem is not _when_ the compilation happens. We can make it an implementation detail, and simply remove the whole concept of "just in time" compilation. But we want to keep backward compatibility as much as possible. We don't want to have people change their workflows just because we compile things differently.

ThomasZitzler:
I don't know the impact but of course we could just ignore the setting of "CompileExpressions" and always treat it as "true". The question is only if that has a performance penalty compared to using the script host.

lbargaoanu:
We're moving away from compilation in the core package. Check #90 and #91. You should be able to bring your own compiler. If you want to contribute what's missing out of the box, a PR is welcome. I'll close this as we're not going to merge it, but feel free to discuss :)

PR #92, opened 2020-03-18
Update copyright for .NET Foundation

dmetzgar:
The LegalCopyright resource string is not used anywhere. I removed all but the default en translation.

@lbargaoanu are we still using the nuspec?


lbargaoanu:
No.

PR #111, opened 2020-04-01
Expired delay with persistence

lbargaoanu:
@dmetzgar I think [we want the null check](https://github.com/UiPath-Open/corewf/blob/master/src/CoreWf/Statements/TimerTable.cs#L149). But writing a failing test is another matter :) That test passes on master, but debugging it, you can see the NRE.


dmetzgar (approved):
Ya, I'm not sure about the validity of that test (the commented out one). If it was trying to do an expression, it needs the brackets.

Your new test works fine for me.

PR #131, opened 2021-01-22
Bringing C# into WF

aoltean16:
### Bringing C# into WF library
- Created all equivalent classes for C# to be supported into WF, same as VB
- Refactored code that resulted as duplicate into C# into more generic classes
- For compatibility reasons, VisualBasicSettings and VisualBasicImportReference do not have a C# equivalent, as agreed with Bogdan Toma
- VBValue and VBReference will now be serialized more explicitly into xaml, the same as its C# equivalent always did.

TODO: 

- [ ] unit tests

WF designer pull request & studio one (in order to test) will come shortly


aoltean16:
Changes here are better reflected in #132

PR #149, opened 2021-05-29
Add AsyncTaskCodeActivity

### LINKED ISSUE #145: Add AsyncTaskCodeActivity type

TheBrambleShark:
Fixes #145

* [x] Define `VoidResult` type (represents a void type since you can't return `System.Void`)
* [x] Define `AsyncTaskCodeActivity` type
* [x] Write unit tests
    * [x] VoidResultTests
    * [x] AsyncTaskCodeActivityTests

Usage Example:
```cs
public class DownloadWebPage : AsyncTaskCodeActivity<string>
{
    [RequiredArgument]
    [OverloadGroup("Address")]
    public InArgument<string> Url { get; set; }

    [RequiredArgument]
    [OverloadGroup("Address")]
    public InArgument<Uri> Uri { get; set; }

    protected override async Task<string> ExecuteAsync
    (
        AsyncCodeActivityContext context,
        CancellationToken cancellationToken
    )
    {
        var url = Url.Get(context);
        var uri = Uri.Get(context);

        var path = url ?? uri.AbsoluteUri;

        return await Download(path, cancellationToken);
    }

    private async Task<string> Download(string path, CancellationToken cancellationToken)
    {
        IFlurlResponse response = await path.GetAsync(cancellationToken);

        switch (response.StatusCode)
        {
            case 200:
                return await response.ResponseMessage.Content.ReadAsStringAsync();
            case 404:
                return "404: Not Found";
            default:
                return $"{response.StatusCode}: {response.ResponseMessage.ReasonPhrase}";
        }
    }
}
```


TheBrambleShark:
PR is complete. Let me know if there are any changes you would like for me to make.

Also, sorry for the delay on this.

src/Test/TestCases.Activities/TerminateWorkflowTests.cs
```diff
@@ -208,7 +208,7 @@ private static void runTerminate()
 
         private class MyTracking : TrackingParticipant
         {
-            protected override void Track(TrackingRecord record, TimeSpan timeout)
+            protected internal override void Track(TrackingRecord record, TimeSpan timeout)
```
TheBrambleShark: Build was erroring with following error:
```
'TerminateWorkflowTests.MyTracking.Track(TrackingRecord, TimeSpan)': cannot change access modifiers when overriding 'protected internal' inherited member 'TrackingParticipant.Track(TrackingRecord, TimeSpan)'
```
Adding `internal` was suggested to match the method it was overriding, so the suggestion was applied.

src/CoreWf/AsyncTaskCodeActivity.cs
```diff
@@ -0,0 +1,154 @@
...
+    {
+        protected sealed override IAsyncResult BeginExecute(AsyncCodeActivityContext context, AsyncCallback callback, object state)
+        {
+            var cts = new CancellationTokenSource();
+            context.UserState = cts;
+
+            var task = ExecuteAsync(context, cts.Token);
+
+            // VoidResult allows us to simulate a void return type.
+            var tcs = new TaskCompletionSource<VoidResult>(state);
```
TheBrambleShark: .NET 5 adds a new non-generic `TaskCompletionSource` to handle exactly this situation. However, the type is not backwards-compatible. I originally had an ifdef to switch between the generic and non-generic version based on the targeted .NET version, however I felt this added unnecessary complexity and opted to just use the generic version.

src/CoreWf/Attributes.cs
```diff
@@ -0,0 +1,3 @@
+﻿using System.Runtime.CompilerServices;
+
+[assembly:InternalsVisibleTo("TestCases.Activities")]
```
TheBrambleShark: `VoidResult` is internal because consumers don't need to be aware of it. As a result, this assembly attribute was necessary to expose `VoidResult` to the testing library.

src/CoreWf/IAsyncTaskCodeActivity.cs
```diff
@@ -0,0 +1,15 @@
+﻿using System.Threading;
+using System.Threading.Tasks;
+
+namespace System.Activities
+{
+    internal interface IAsyncTaskCodeActivity : IAsyncCodeActivity
```
TheBrambleShark: These interfaces are only used in one place. I added them for consistency with the `AsyncCodeActivity` type, however they could easily be done away with if we feel it adds bloat.

src/CoreWf/VoidResult.cs
```diff
@@ -0,0 +1,57 @@
+﻿using System.Threading.Tasks;
+
+namespace System.Activities
+{
+    /// <summary>
+    /// Represents a void type, since <see cref="System.Void"/> is not a valid return type in C#.
+    /// </summary>
+    /// <remarks>Based on the Unit type from MediatR.</remarks>
+    internal readonly struct VoidResult : IEquatable<VoidResult>, IComparable<VoidResult>, IComparable
```
TheBrambleShark: `VoidResult` is based on the [Unit](https://github.com/jbogard/MediatR/blob/master/src/MediatR/Unit.cs) type from MediatR, licensed under the [Apache 2.0](https://github.com/jbogard/MediatR/blob/master/LICENSE) license.

The only change made to the type was its renaming to `VoidResult`, which I felt better showed its intent.

src/CoreWf/IAsyncTaskCodeActivity.cs
lbargaoanu: Yes, I don't think we need these interfaces.

src/CoreWf/VoidResult.cs
lbargaoanu: I understand the idea, but given the fact that we don't really use much of this `void` type, I think we can simply use `object` instead. It's just an implementation detail, we can always introduce such a type if the need arises.

src/Test/TestCases.Activities/TerminateWorkflowTests.cs
lbargaoanu: The project builds and the CI proves that :) Just revert this unrelated change.

src/Test/TestCases.Activities/AsyncTaskCodeActivityTests.cs
```diff
@@ -0,0 +1,42 @@
+﻿using System.Activities;
+using System.Threading.Tasks;
+using TestObjects.CustomActivities;
+using Xunit;
+
+namespace TestCases.Activities
+{
+    public sealed class AsyncTaskCodeActivityTests
+    {
+        private static readonly AsyncTaskActivity<object> genericActivity = new(Task.FromResult<object>(null));
```
lbargaoanu: This is only used in one test, so it should be declared there.
Also, I don't see any test for the non generic version.

src/CoreWf/Attributes.cs
TheBrambleShark: Removing as it's no longer needed - VoidResult doesn't exist anymore.

lbargaoanu: Revert?

PR #172, opened 2021-12-05
Convert to .NET 6, made style consistent in System.Activities [UI-52330]

dmetzgar:
Made the following changes:
- Updated to .NET 6
- Converted to file-scoped namespaces
- Added global usings for common namespaces
- Non-static fields prefixed with underscore, removed `this.`
- Used expression body for properties, accessors, and methods where applicable
- Followed suggested fixes from IDE
- Renamed `#if` regions specific to dynamic update feature to `#if DYNAMICUPDATE`. There's no intent to support dynamic update so this makes it easier to remove.
- Removed dead code
- Split out the inner classes of some of the larger (>1000 line) files

No public surface area was changed. All tests pass.

Also added a Performance test using Benchmark .NET. It only contains one test so far to validate a XAML with 400 expressions. Results before and after moving to .NET 6 are below. Validate calls CacheMetadata on each activity in the workflow. For VB each expression is compiled individually. For C#, a `CompiledExpressionInvoker` is created but the expression compilation is not performed. In the .NET Framework, all the C# expressions were compiled at the end using MSBuild. This is why SkipExpressionCompilation, which uses VB, and CSharp test cases below have the same speed.

## Targeting .NET 5

No changes to dependencies. Using .NET 6 SDK.

```
BenchmarkDotNet=v0.13.1, OS=Windows 10.0.19043.1348 (21H1/May2021Update)
Intel Core i7-8700 CPU 3.20GHz (Coffee Lake), 1 CPU, 12 logical and 6 physical cores
.NET SDK=6.0.100
  [Host]     : .NET 5.0.12 (5.0.1221.52207), X64 RyuJIT
  DefaultJob : .NET 5.0.12 (5.0.1221.52207), X64 RyuJIT
```

|                    Method |         Mean |       Error |      StdDev |    Ratio | RatioSD |
|-------------------------- |-------------:|------------:|------------:|---------:|--------:|
| SkipExpressionCompilation |     3.423 ms |   0.0073 ms |   0.0065 ms |     1.00 |    0.00 |
|                        VB | 8,251.088 ms | 142.3681 ms | 133.1712 ms | 2,404.58 |   33.59 |
|                    CSharp |     3.385 ms |   0.0117 ms |   0.0110 ms |     0.99 |    0.00 |

## Targeting .NET 6

No changes to dependencies. Using .NET 6 SDK.

```
BenchmarkDotNet=v0.13.1, OS=Windows 10.0.19043.1348 (21H1/May2021Update)
Intel Core i7-8700 CPU 3.20GHz (Coffee Lake), 1 CPU, 12 logical and 6 physical cores
.NET SDK=6.0.100
  [Host]     : .NET 6.0.0 (6.0.21.52210), X64 RyuJIT
  DefaultJob : .NET 6.0.0 (6.0.21.52210), X64 RyuJIT
```

|                    Method |         Mean |      Error |     StdDev |    Ratio | RatioSD |
|-------------------------- |-------------:|-----------:|-----------:|---------:|--------:|
| SkipExpressionCompilation |     5.117 ms |  0.0164 ms |  0.0137 ms |     1.00 |    0.00 |
|                        VB | 6,808.368 ms | 48.6314 ms | 43.1104 ms | 1,330.04 |    8.30 |
|                    CSharp |     5.098 ms |  0.0205 ms |  0.0171 ms |     1.00 |    0.00 |


lbargaoanu:
We don't need anything in .Net 6 at this time, so no actual benefit, and the downside is that I'm forced to use VS 2022.

lbargaoanu:
Maybe I'm missing smth, but the first test runs on .Net 5 and the second on .Net 6. But you can always run the version targeting .Net 5 on .Net 6. And I wouldn't expect much of a difference then.

PR #178, opened 2022-01-13
VB expression validation performance improvements [UI-50382]

dmetzgar:
`ActivityValidationServices.Validate` is used by workflow designer applications to get a list of validation errors that can be presented to the user. Expressions are an important part of this because the user should know if an expression is invalid. If the workflow has a lot of Visual Basic expressions, the validate step will take a long time. For example, here is a BenchmarkDotNet result of validating a workflow with 400 expressions:

|                  Method |        Mean |     Error |      StdDev |
|------------------------ |------------:|----------:|------------:|
|              VB400Stmts | 20,694.6 ms | 728.41 ms | 2,101.62 ms |

A PerfView profile reveals that most of the time spent in ActivityValidationServices.Validate is in building an assembly for each expression in the workflow:

| Name                                                                                                                                                                     |  Inc % |
|--------------------------------------------------------------------------------------------------------------------------------------------------------------------------|-------:|
| + System.Activities!ActivityValidationServices.Validate                                                                                                                  |  100.0 |
|  + System.Activities!ActivityValidationServices.Validate                                                                                                                 |  100.0 |
|   + System.Activities!System.Activities.Validation.ActivityValidationServices+InternalActivityValidationServices.InternalValidate()                                      |  100.0 |
|    + System.Activities!ActivityUtilities.CacheRootMetadata                                                                                                               |  100.0 |
|     + System.Activities!ActivityUtilities.ProcessActivityTreeCore                                                                                                        |  100.0 |
|      + System.Activities!ActivityUtilities.ProcessActivity                                                                                                               |  100.0 |
|       + System.Activities!Activity.InternalCacheMetadata                                                                                                                 |   99.9 |
|       \|+ System.Activities!System.Activities.Activity\`1[System.__Canon].OnInternalCacheMetadata(bool)                                                                  |   99.9 |
|       \| + System.Activities!System.Activities.CodeActivity\`1[System.__Canon].OnInternalCacheMetadataExceptResult(bool)                                                 |   99.9 |
|       \|  + uipath.workflow!Microsoft.VisualBasic.Activities.VisualBasicValue\`1[System.__Canon].CacheMetadata(value class System.Activities.CodeActivityMetadata)       |   99.9 |
|       \|   + uipath.workflow!VisualBasicHelper.Compile                                                                                                                   |   99.9 |
|       \|    + uipath.workflow!System.Activities.JitCompilerHelper\`1[System.__Canon].Compile(value class System.Activities.CodeActivityPublicEnvironmentAccessor,bool)   |   99.9 |
|       \|     + uipath.workflow!System.Activities.JitCompilerHelper\`1[System.__Canon].Compile(class System.Activities.LocationReferenceEnvironment,bool)                 |   99.9 |
|       \|      + uipath.workflow!ScriptingJitCompiler.CompileExpression                                                                                                   |   99.8 |
|       \|      \|+ uipath.workflow!ScriptingAotCompiler.BuildAssembly                                                                                                     |   89.0 |

89% of the cost of Validate is in BuildAssembly. The code for BuildAssembly is as follows:

```csharp
public static TextExpressionCompilerResults BuildAssembly(Compilation compilation, string typeName, Stream copy = null)
{
    var results = GetCompilerResults(compilation);
    if (results.HasErrors)
    {
        return results;
    }
    using var stream = new MemoryStream();
    var emitResult = compilation.Emit(stream);
    AddDiagnostics(results, emitResult.Diagnostics);
    if (!emitResult.Success)
    {
        return results;
    }
    results.ResultType = Assembly.Load(stream.GetBuffer()).GetType(typeName, throwOnError: true);
    if (copy != null)
    {
        stream.Position = 0;
        stream.CopyTo(copy);
    }
    return results;
}
```

Diagnostics are retrieved from the Compilation object. The Compilation object comes from the Script object created by VisualBasicScript. My initial suspicion was that emitting the Assembly to a memory stream and then loading it was the primary cause of the slow performance. But skipping the emit and load only reduces the duration of validating the 400 expression workflow by half.

|                  Method |    Mean |    Error |   StdDev |
|------------------------ |--------:|---------:|---------:|
| VB400Stmts_ValidateOnly | 9.907 s | 0.1182 s | 0.1105 s |

A significant portion of the performance issue comes from the Script implementation itself. An odd thing about the Script implementation for VB is that it was dropped from Roslyn: https://github.com/dotnet/roslyn/issues/13523#issuecomment-462433207

The code for the Microsoft.CodeAnalysis.VisualBasic.Scripting library was copied from the Roslyn repository and added to CoreWF. This was necessary because the library makes calls to internals in the Microsoft.CodeAnalysis.Scripting library, which has an `InternalsVisibleTo` that allows Microsoft.CodeAnalysis.VisualBasic.Scripting access (see https://github.com/dotnet/roslyn/blob/main/src/Scripting/Core/Microsoft.CodeAnalysis.Scripting.csproj#L52). Luckily, the InternalsVisibleTo only specifies the Assembly short name and no signing key.

The situation is a bit fragile and it would be preferable to have a workaround for the VB.Scripting library. Unfortunately, the only ways to fix this are to convince the Roslyn team to make changes in their libraries by adding more public surface area or copying all of the VB code. The VisualBasicCompilation class, which is part of the Microsoft.CodeAnalysis.VisualBasic assembly that is still shipped and supported by Roslyn, creates a few classes when `Emit` is called that are internal only. 

What this means is that if I create a VisualBasicCompilation object and call Emit, it will fail. The internal classes are not accessible because the generated code was not created by a friend assembly like Microsoft.CodeAnalysis.VisualBasic.Scripting. The class that causes the failure is here: https://github.com/dotnet/roslyn/blob/315c2e149ba7889b0937d872274c33fcbfe9af5f/src/Compilers/VisualBasic/Portable/Symbols/EmbeddedSymbols/VbCoreSourceText.vb#L1063.

The emitted code is explained here: https://github.com/dotnet/roslyn/blob/315c2e149ba7889b0937d872274c33fcbfe9af5f/src/Compilers/VisualBasic/Portable/Lowering/LocalRewriter/LocalRewriter_UnstructuredExceptionHandling.vb#L79

This necessitates two paths in the code. One path is for compiling the expression since only emitting and loading will give access to the LambdaExpression that is needed to execute the expression at runtime. The other path is for expression validation. Validation only requires the diagnostic information, which has compiler messages like you would see if you put the expression in the Immediate Window in Visual Studio. Diagnostic information can be obtained without calling Emit so the VisualBasicScript is not needed.

But it's not easy to tell when the VisualBasicValue is being validated in a designer versus being executed at runtime. This stems from the overuse of the `CacheMetadata` method on the activity. `ActivityValidationServices.Validate` really just calls CacheMetadata and returns the ValidationErrors. CacheMetadata was originally intended for use at runtime to --cache metadata-- but the cache portion has lost meaning over time. The best way to communicate with expression activities is to put information in the [LocationReferenceEnvironment](https://docs.microsoft.com/en-us/dotnet/api/system.activities.locationreferenceenvironment?view=netframework-4.8).

One of the difficulties with the LocationReferenceEnvironment is the use of the `CompileExpressions` flag. When this flag is set, expression activities should not attempt to validate the expression. This is a strange behavior that came about in .NET 4.5 when WF added support for C# expressions. The VB expressions supported in .NET 4.0 are compiled using a custom-built native VB compiler made exclusively for WF. There was no such equivalent for C# in .NET 4.5 except to call MSBuild to invoke the csc compiler. So workflows with C# expressions would emit types with the expressions in them and use custom MSBuild tasks to compile them. The workflow designer in Visual Studio would validate individual C# expressions at design time using a service built into Visual Studio. With Roslyn, these different approaches are unnecessary (forgetting the Script class issues mentioned earlier) and CoreWF could have a unified approach to both languages (not in this PR though).

This PR skips VisualBasicScript for the validation path and creates a VisualBasicCompilation object instead. It includes some tests to show the expression validation does catch invalid expressions. More extensive testing is needed to determine if there are any functional gaps in the diagnostics. Performance tests show a significant improvement with the 400 expression workflow validation.

|                  Method |        Mean |     Error |      StdDev |
|------------------------ |------------:|----------:|------------:|
| VB400Stmts (Before) | 20,694.6 ms | 728.41 ms | 2,101.62 ms |
| VB400Stmts (After) |    116.1 ms |   2.30 ms |     4.85 ms |

The full benchmark results for expressions now looks like:

```
BenchmarkDotNet=v0.13.1, OS=Windows 10.0.19043.1415 (21H1/May2021Update)
Intel Core i7-8700 CPU 3.20GHz (Coffee Lake), 1 CPU, 12 logical and 6 physical cores
.NET SDK=6.0.101
  [Host]     : .NET 6.0.1 (6.0.121.56705), X64 RyuJIT
  DefaultJob : .NET 6.0.1 (6.0.121.56705), X64 RyuJIT
```

|          Method |         Mean |       Error |      StdDev |       Median |
|---------------- |-------------:|------------:|------------:|-------------:|
|      VB100Stmts |  35,081.6 us | 1,386.51 us | 3,818.86 us |  33,626.3 us |
|      VB400Stmts | 123,931.2 us | 1,549.09 us | 1,449.02 us | 123,461.4 us |
| VBSingleExpr100 |     595.9 us |     3.77 us |     3.34 us |     594.7 us |

### DIFF
--- a/src/Perf/CoreWf.Benchmarks/Expressions.cs
+++ b/src/Perf/CoreWf.Benchmarks/Expressions.cs
@@ -1,4 +1,5 @@
 ﻿using BenchmarkDotNet.Attributes;
+using Microsoft.CSharp.Activities;
 using Microsoft.VisualBasic.Activities;
 using System.Activities;
 using System.Activities.Statements;
@@ -13,6 +14,7 @@ public class Expressions
     private readonly Activity[] _vb100Stmts;
     private readonly Activity[] _vb400Stmts;
     private readonly Activity[] _vbSingleExpr100;
+    private readonly Activity[] _cs400Stmts;
     private int _activityIndex;
 
     public Expressions()
@@ -37,6 +39,12 @@ public class Expressions
             GenerateManyExpressionsWorkflow(800, 400),
         };
         _vbSingleExpr100 = GenerateManySingleExpressionWorkflows(500, 100);
+        _cs400Stmts = new Activity[]
+        {
+            GenerateManyExpressionsWorkflow(0, 400, "CS"),
+            GenerateManyExpressionsWorkflow(400, 400, "CS"),
+            GenerateManyExpressionsWorkflow(800, 400, "CS"),
+        };
         _activityIndex = 0;
     }
 
@@ -96,15 +104,31 @@ public class Expressions
         ActivityXamlServices.Compile(root, new());
     }
 
-    private static Activity GenerateManyExpressionsWorkflow(int startExprNum, int numExpressions)
+    [Benchmark]
+    public void CS400Stmts()
+    {
+        var activity = _cs400Stmts[_activityIndex];
+        _activityIndex = (_activityIndex + 1) % _cs400Stmts.Length;
+        _ = ActivityValidationServices.Validate(activity);
+    }
+
+    private static Activity GenerateManyExpressionsWorkflow(int startExprNum, int numExpressions, string language = "VB")
     {
         Sequence workflow = new();
         int endExprNum = startExprNum + numExpressions;
         for (int i = startExprNum; i < endExprNum; i++)
         {
-            VisualBasicValue<string> vbv = new(@$"String.Concat(""alpha "", ""beta "", {i})");
             WriteLine writeLine = new();
-            writeLine.Text = new InArgument<string>(vbv);
+            if (language == "VB")
+            {
+                VisualBasicValue<string> vbv = new(@$"String.Concat(""alpha "", ""beta "", {i})");
+                writeLine.Text = new InArgument<string>(vbv);
+            }
+            else
+            {
+                CSharpValue<string> csv = new($@"string.Join(',', ""alpha"", ""beta"", {i})");
+                writeLine.Text = new InArgument<string>(csv);
+            }
             workflow.Activities.Add(writeLine);
         }
 
--- a/src/Perf/CoreWf.Benchmarks/Expressions.cs
+++ b/src/Perf/CoreWf.Benchmarks/Expressions.cs
@@ -3,6 +3,7 @@ using Microsoft.VisualBasic.Activities;
 using System.Activities;
 using System.Activities.Statements;
 using System.Activities.Validation;
+using System.Activities.XamlIntegration;
 using System.Text;
 
 namespace CoreWf.Benchmarks;
@@ -39,7 +40,7 @@ public class Expressions
         _activityIndex = 0;
     }
 
-    [Benchmark]
+    //[Benchmark]
     public void VB100Stmts()
     {
         var activity = _vb100Stmts[_activityIndex];
@@ -55,12 +56,44 @@ public class Expressions
         _ = ActivityValidationServices.Validate(activity);
     }
 
-    [Benchmark]
+    //[Benchmark]
     public void VBSingleExpr100()
     {
         var activity = _vbSingleExpr100[_activityIndex];
         _activityIndex = (_activityIndex + 1) % _vbSingleExpr100.Length;
         _ = ActivityValidationServices.Validate(activity);
+    }
+
+    [Benchmark]
+    public void VB400Stmts_AOT()
+    {
+#if RELEASE
+        var workflow = _vb400Stmts[_activityIndex];
+        _activityIndex = (_activityIndex + 1) % _vb400Stmts.Length;
+#else
+        Sequence workflow = new();
+        for (int i = 0; i < 200; i++)
+        {
+            VisualBasicValue<string> vbv = new(@$"String.Concat(""alpha "", ""beta "", {i})");
+            WriteLine writeLine = new();
+            writeLine.Text = new InArgument<string>(vbv);
+            workflow.Activities.Add(writeLine);
+        }
+        VisualBasicValue<string> badVbv = new(@$"String.Concat(""alpha "", b, ""beta "", 200)");
+        WriteLine badWriteLine = new();
+        badWriteLine.Text = new InArgument<string>(badVbv);
+        workflow.Activities.Add(badWriteLine);
+        for (int i = 201; i < 400; i++)
+        {
+            VisualBasicValue<string> vbv = new(@$"String.Concat(""alpha "", ""beta "", {i})");
+            WriteLine writeLine = new();
+            writeLine.Text = new InArgument<string>(vbv);
+            workflow.Activities.Add(writeLine);
+        }
+#endif
+
+        var root = new DynamicActivity { Implementation = () => workflow };
+        ActivityXamlServices.Compile(root, new());
     }
 
     private static Activity GenerateManyExpressionsWorkflow(int startExprNum, int numExpressions)
--- a/src/Perf/CoreWf.Benchmarks/Program.cs
+++ b/src/Perf/CoreWf.Benchmarks/Program.cs
@@ -10,7 +10,8 @@ try
     //e.CSharp();
     //e.VB();
     //e.VBSingleExpr100();
-    e.VB400Stmts();
+    //e.VB400Stmts();
+    e.CS400Stmts();
     //e.VBSingleExpr100Stmts();
     //e.VBSingleExpr200Stmts();
     //e.VBSingleExpr400Stmts();
--- a/src/Perf/CoreWf.Benchmarks/Program.cs
+++ b/src/Perf/CoreWf.Benchmarks/Program.cs
@@ -9,11 +9,12 @@ try
 {
     //e.CSharp();
     //e.VB();
-    e.VBSingleExpr100();
-    //e.VB400Stmts();
+    //e.VBSingleExpr100();
+    e.VB400Stmts();
     //e.VBSingleExpr100Stmts();
     //e.VBSingleExpr200Stmts();
     //e.VBSingleExpr400Stmts();
+    //e.VB400Stmts_AOT();
 }
 catch (Exception ex)
 {
--- a/src/Perf/CoreWf.Benchmarks/VbCompilation.cs
+++ b/src/Perf/CoreWf.Benchmarks/VbCompilation.cs
@@ -1,0 +1,48 @@
+﻿using BenchmarkDotNet.Attributes;
+using Microsoft.CodeAnalysis;
+using Microsoft.CodeAnalysis.VisualBasic;
+
+namespace CoreWf.Benchmarks;
+
+public class VbCompilation
+{
+    private static int GlobalCounter = 0;
+    private VisualBasicCompilation _visualBasicCompilation;
+    private VisualBasicParseOptions _parseOptions = new(kind: SourceCodeKind.Script);
+
+    public VbCompilation()
+    {
+        string assemblyName = Guid.NewGuid().ToString();
+        List<MetadataReference> references = new();
+        MetadataReference mr = MetadataReference.CreateFromFile(typeof(string).Assembly.Location);
+        references.Add(mr);
+        VisualBasicCompilationOptions options = new(
+            OutputKind.DynamicallyLinkedLibrary);
+        _visualBasicCompilation = VisualBasicCompilation.Create(assemblyName, /*syntax trees*/ null, references, options);
+        var syntaxTree = VisualBasicSyntaxTree.ParseText(GetExpressionText(), _parseOptions);
+        _visualBasicCompilation = _visualBasicCompilation.AddSyntaxTrees(syntaxTree);
+        var diagnostics = _visualBasicCompilation.GetDiagnostics();
+        if (diagnostics.Any(d => d.Severity >= DiagnosticSeverity.Error))
+        {
+            Console.WriteLine("Error encountered");
+        }
+    }
+
+    [Benchmark]
+    public void MetadataReferenceCreate()
+    {
+        _ = MetadataReference.CreateFromFile(typeof(string).Assembly.Location);
+    }
+
+    [Benchmark]
+    public void SingleExpression()
+    {
+        var oldSyntaxTree  = _visualBasicCompilation.SyntaxTrees[0];
+        var syntaxTree = VisualBasicSyntaxTree.ParseText(GetExpressionText(), _parseOptions);
+        _visualBasicCompilation = _visualBasicCompilation.ReplaceSyntaxTree(oldSyntaxTree, syntaxTree);
+        _ = _visualBasicCompilation.GetDiagnostics();
+    }
+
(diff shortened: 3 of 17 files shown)


lbargaoanu:
But why bother with the JIT compiler for Validate, why not use the AOT compiler, which creates a single assembly?

dmetzgar:
> But why bother with the JIT compiler for Validate, why not use the AOT compiler, which creates a single assembly?

Could you show me how that would work?

lbargaoanu:
There are passing tests that use the AOT compiler, you just set `CompileExpressions` to `true`.

dmetzgar:
> There are passing tests that use the AOT compiler, you just set `CompileExpressions` to `true`.

Thanks for pointing this out. Compiling everything in one shot is faster than compiling each expression individually but I was unsure how we could tie an expression error back to the activity that contains the expression. For example, if I change one of the 400 expressions to be invalid, I get the following error message on AOT compile.

```
Compilation failures occurred: 
 Line 12856: (12857) : error BC30451: 'b' is not declared. It may be inaccessible due to its protection level.
 Line 12862: (12863) : error BC30451: 'b' is not declared. It may be inaccessible due to its protection level.
Complete results are contained in the Data property of this exception.  Please correct the errors in the source and retry the Load.
```

I need a way of correlating this compiler error back to the expression. Unfortunately, the Data field in the exception doesn't contain anything else and there's nothing indicating the name of the function that houses the expression. The generated source code around that line number is below. The original invalid expression is `String.Concat("alpha ", b, "beta ", 200)`, which is invalid because `b` is not defined.

```vbnet
        Friend Function __Expr200GetTree() As System.Linq.Expressions.Expression
            Dim expression As System.Linq.Expressions.Expression(Of System.Func(Of String)) = Function() String.Concat("alpha ", b, "beta ", 200)
            Return MyBase.RewriteExpressionTree(expression)
        End Function
        
        <System.Diagnostics.DebuggerHiddenAttribute()>  _
        Public Function __Expr200Get() As String
            Return String.Concat("alpha ", b, "beta ", 200)
        End Function
        
        Public Function ValueType___Expr200Get() As String
            Me.GetValueTypeValues
            Return Me.__Expr200Get
        End Function
```

What I'm unsure of is if the function name is just an incrementing counter of expressions or if it's derived from the InternalId of the VisualBasicValue activity. If the latter, I could track the method name back to the activity and put the validation errors in that activity's metadata. The full code is available at the time the compiler errors are processed. This is the Compile method from the ScriptingAotCompiler:

```csharp
public abstract class ScriptingAotCompiler : AheadOfTimeCompiler
{
    protected abstract Script<object> Create(string code, ScriptOptions options);
    public override TextExpressionCompilerResults Compile(ClassToCompile classToCompile)
    {
        //File.WriteAllText(@"AotSource.vb", classToCompile.Code);
        var scriptOptions = ScriptOptions.Default
            .WithReferences(classToCompile.ReferencedAssemblies.GetMetadataReferences())
            .WithImports(classToCompile.ImportedNamespaces)
            .WithOptimizationLevel(OptimizationLevel.Release);
        var script = Create(classToCompile.Code, scriptOptions);
        var results = BuildAssembly(script.GetCompilation());
        if (results.HasErrors)
        {
            return results;
        }
        results.ResultType = results.ResultType.GetNestedType(classToCompile.ClassName);
        return results;
    }
```

I could use the line numbers to track within `classToCompile.Code` to get the function. Maybe I could emit some extra lines like comments that I could use to track the expression back to the right activity. Having some kind of debug information would be awesome. Unfortunately, setting the OptimizationLevel to Debug doesn't change the error message or information in the results.

I modified the code a little bit to see if I could get more information from the diagnostics:

```csharp
        var script = Create(classToCompile.Code, scriptOptions);
        var compilation = script.GetCompilation();
        var diagnostics = compilation.GetDiagnostics();
```

The Compilation object generated from the script has a single SyntaxTree in it and diagnostics don't seem to reveal the function where the compiler error is found. The best I can do is the line number. It's possible to use AOT by adding some cleverness around looking back into the generated code and getting the InternalId of the activity, then adding the validation errors to that activity's metadata. It could be a quality of life improvement as well if expression issues are not caught in the designer - like if the workflow was generated, so I think it's an avenue worth exploring.

Assuming I was able to get the compiler errors to correlate back correctly, we can benchmark that performance against the method proposed in this PR. The code for the AOT benchmark follows. I'll add this to the benchmarks as well.

```csharp
    public void VB400Stmts_AOT()
    {
        var workflow = _vb400Stmts[_activityIndex];
        _activityIndex = (_activityIndex + 1) % _vb400Stmts.Length;
        var root = new DynamicActivity { Implementation = () => workflow };
        ActivityXamlServices.Compile(root, new());
    }
```

Here's the resulting benchmark compared against the JIT method:

|          Method |         Mean |       Error |      StdDev |
|---------------- |-------------:|------------:|------------:|
|  VB400Stmts_JIT | 122,632.9 us | 1,841.57 us | 1,632.50 us |
|  VB400Stmts_AOT | 830,429.2 us | 6,638.75 us | 6,209.89 us |
|  VB400Stmts_AOT_DiagnosticsOnly | 364,988.3 us | 7,072.32 us | 8,144.50 us |

Granted, AOT is significantly faster than the original JIT code but still a few times slower than the new JIT validation code. The "DiagnosticsOnly" run is if we use a similar approach as to what's done in this PR where a separate path just gets the diagnostics without calling BuildAssembly. This was just a hack in the code to remove the BuildAssembly portion to see how it would work. I'm surprised it's still slower than the JIT version and I haven't done a profile to find out why.

lbargaoanu:
I don't know enough about when validating _all_ expressions is required. I kind of assumed you start from a valid workflow, validate a single expression and get to another valid workflow. And you validate everything when you just want to compile all expressions. Clearly there are issues with namespaces, references, variables that need special consideration. While I hope validating everything is a rare occurence, I would like to talk with the guys from Studio as I expect they're in the know. But I agree that tracking back to the activity is tricky (ideas: `#line` and `diagnostic.Location`).
One reason the new code compiles faster is that it does smth else. There is no expression to speak of :) And the genereted AOT code is not pretty.
Implementation aside, I think this diagnostics only mode should be separated from the existing code. Maybe another compiler or another API. I'd rather not add ad-hoc flags and change existing API, although I'm guilty of that too :)
But whatever we do, the affected people should be part of the conversation.

lbargaoanu:
Maybe a technicality, but not all VB/C# can be translated to expressions, so validating this way is a subset of the runtime validations. Some valid expressions will fail at runtime. Granted, that sometimes happens even on 461.

lbargaoanu:
Why not write this logic in a custom JIT compiler? Then you don't need the new abstractions and the new flags. You can set the new compiler before validating and set the default back when it's done.

dmetzgar:
> Why not write this logic in a custom JIT compiler? Then you don't need the new abstractions and the new flags. You can set the new compiler before validating and set the default back when it's done.

The JIT compiler requires creating a LambdaExpression, which I don't think is necessary for validating an expression. 

The earlier point about the diagnostics being a subset of runtime validations is something that doesn't apply at design time, where the LambdaExpression is unused. The designer wants to show if there are validation errors with the expression and those are only obtained by throwing an exception from the compiler.

The JitCompiler uses AotCompiler.BuildAssembly to build the assembly ([here](https://github.com/UiPath/CoreWF/blob/c503f51982a5fc396453571ec6f3fdd31b59c06c/src/UiPath.Workflow/Activities/ScriptingJitCompiler.cs#L65)), get the generated type, and execute a method that returns a LambdaExpression with the original expression in it. The original expression itself isn't executed.
```csharp
var results = ScriptingAotCompiler.BuildAssembly(finalCompilation);
if (results.HasErrors)
{
    throw FxTrace.Exception.AsError(new SourceExpressionException(SR.CompilerErrorSpecificExpression(expressionToCompile.Code, results), results.CompilerMessages));
}
return (LambdaExpression)results.ResultType.GetMethod("CreateExpression").Invoke(null, null);
```

Comparing to BuildAssembly, the main difference in my approach is that it stops before the Emit.
```csharp
public static TextExpressionCompilerResults BuildAssembly(Compilation compilation, string typeName, Stream copy = null)
{
    var results = GetCompilerResults(compilation);
    if (results.HasErrors)
    {
        return results;
    }
    using var stream = new MemoryStream();
    var emitResult = compilation.Emit(stream);
    AddDiagnostics(results, emitResult.Diagnostics);
    if (!emitResult.Success)
    {
        return results;
    }
    results.ResultType = Assembly.Load(stream.GetBuffer()).GetType(typeName, throwOnError: true);
    if (copy != null)
    {
        stream.Position = 0;
        stream.CopyTo(copy);
    }
    return results;
}
```

The validator is basically the equivalent of the GetCompilerResults. The Emit has diagnostics in its results so it begs the question: what would be caught in Emit that is not caught through GetCompilerResults? My answer to that is that there is no difference for a single expression. The expression can only interact with the parameters passed into the function (determined by the scope of the activity) and with types that are available via imported references. All of which should be checked by Compilation.GetDiagnostics().

The other difference in my approach is not using the Compilation object obtained from the Microsoft.CodeAnalysis.VisualBasic.Scripting library, which adds [extra code](https://github.com/dotnet/roslyn/blob/315c2e149ba7889b0937d872274c33fcbfe9af5f/src/Compilers/VisualBasic/Portable/Lowering/LocalRewriter/LocalRewriter_UnstructuredExceptionHandling.vb#L79) during the Emit that is not available in the vanilla Compilation object used in this PR. This extra step is slow and I don't see how it helps in determining the validity of the expression.

lbargaoanu:
The idea was to return some bogus `LambdaExpression` :) Have the same code, just the compiler interface.
Not having expressions clearly declares valid expressions that will fail at runtime. Your code compiles when some text can be transtated to IL and sometimes that code cannot be translated to expressions.
Maybe that's even acceptable, but we have to be explicit about it.

lbargaoanu:
https://github.com/dotnet/csharplang/discussions/158

dmetzgar:
> [dotnet/csharplang#158](https://github.com/dotnet/csharplang/discussions/158)

This is a good catch. The report of the compilation errors with lambda expressions when running the AOT compiler actually comes from the GetDiagnostics step before the Emit. My code wasn't catching it because it wasn't creating the lambda expression. Fixing that, it now catches the null conditional.

Benchmark results show it's still roughly the same speed as before:

|         Method |     Mean |    Error |   StdDev |
|--------------- |---------:|---------:|---------:|
|     VB400Stmts | 178.6 ms |  1.72 ms |  1.61 ms |
| VB400Stmts_AOT | 859.9 ms | 12.25 ms | 10.86 ms |
|     CS400Stmts | 205.9 ms |  4.08 ms |  7.66 ms |

lbargaoanu:
As long as there's only one client, I think this implementation should live in a custom compiler outside CoreWf.

PR #191, opened 2022-02-16
Visual Basic JIT Expression Compiler

valpiele:
This PR is addressing a problem related to visual basic expressions compilation.
Expression identifiers are compared with workflow variables, which can lead to declaring bad parameters for the generated lambda expression.
Example: 
`Excel.Sheet("sheet").Range(range)` is transformed into:
`Public Shared Function CreateExpression() As Expression(Of Func(Of UiPath.Excel.IWorkbookQuickHandle, String, String, UiPath.Excel.IReadRangeRef))
     Return Function(Excel, Range, range) (Excel.Sheet("sheet").Range(Range))
End Function`

The initial expression contains 4 IdentifierNames, where `Excel` and `range` are the only variables. The matching algorithm (see `ExpressionToCompile.VariableTypeGetter`) will consider the `Range` property to be a variable, which is not true, and will declare it as a parameter. 
A validation error will be visible in the designer because now we're passing 2 parameters with the same name.
Please see the tests for more details.

A possible solution, which may not be the best, is avoiding identifiers that are certainly not variables.

### DIFF
--- a/src/UiPath.Workflow/Activities/ScriptingJitCompiler.cs
+++ b/src/UiPath.Workflow/Activities/ScriptingJitCompiler.cs
@@ -40,6 +40,7 @@ public abstract class ScriptingJitCompiler : JustInTimeCompiler
 
     protected MetadataReference[] MetadataReferences { get; set; }
     protected abstract int IdentifierKind { get; }
+    protected virtual StringComparer IdentifierNameComparer => StringComparer.Ordinal;
     protected abstract string CreateExpressionCode(string types, string names, string code);
     protected abstract string GetTypeName(Type type);
     protected abstract Script<object> Create(string code, ScriptOptions options);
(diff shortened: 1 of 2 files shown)


src/UiPath.Workflow/Activities/ScriptingJitCompiler.cs
```diff
@@ -118,6 +119,23 @@ public class VbJitCompiler : ScriptingJitCompiler
...
+    {
+        return syntaxTree.GetRoot().DescendantNodesAndSelf().Where(n => ValidSyntaxNode(n))
+                         .Select(n => n.ToString()).Distinct().ToArray();
+
+        bool ValidSyntaxNode(SyntaxNode node)
+        {
+            if (node.RawKind != IdentifierKind)
+                return false;
+
+            if (node.Parent.RawKind == (int)SyntaxKind.SimpleMemberAccessExpression && (node.Parent as MemberAccessExpressionSyntax)?.Name?.ToString() == node.ToString())
```
dmetzgar: This looks like a case-sensitive comparison. How is it working?

valpiele: It seems that it is enough to make sure of the uniqueness of the parameters. Please see the latest change. :)

dmetzgar (approved):

PR #196, opened 2022-03-15
Refactor validator, fix expression issues [UI-54756] [UI-55419] [UI-53852]

dmetzgar:
Added an ExpressionContainer class to hold the data on an expression on its way through validation. This allows the extension points to get access to the activity, LRE, imported namespaces, referenced assemblies, etc. and make changes to these at extension points in the validation process.

Also fixed a few bugs. Dictionary access would not work without the System.Collections assembly included, which is due to newer versions of .NET splitting out namespaces from what used to be mscorlib. Linq extensions are also included by default. VB overflow checking is now turned out so that if, for example, a constant expression returns a value that would overflow an integer there will be a validation error.

### DIFF
--- a/src/UiPath.Workflow/Activities/CsExpressionValidator.cs
+++ b/src/UiPath.Workflow/Activities/CsExpressionValidator.cs
@@ -33,6 +33,8 @@ public class CsExpressionValidator : RoslynExpressionValidator
         typeof(CSharpValue<>).Assembly,
     };
 
+    private Compilation DefaultCompilationUnit;
+
     /// <summary>
     ///     Singleton instance of the default validator.
     /// </summary>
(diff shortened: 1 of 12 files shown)


src/UiPath.Workflow/Activities/CsExpressionValidator.cs
```diff
@@ -61,37 +63,45 @@ public CsExpressionValidator(HashSet<Assembly> referencedAssemblies)
...
+        if (DefaultCompilationUnit == null)
         {
             var assemblyName = Guid.NewGuid().ToString();
             CSharpCompilationOptions options = new(
                 OutputKind.DynamicallyLinkedLibrary,
                 mainTypeName: null,
-                usings: expressionToValidate.ImportedNamespaces,
+                usings: expressionContainer.ExpressionToValidate.ImportedNamespaces,
                 optimizationLevel: OptimizationLevel.Debug,
                 checkOverflow: false,
```
liviu-seniuc, line 78: I think `checkOverflow` should be on `true` as for VB

PR #199, opened 2022-03-21
fix: expressions compilation failure will bubble up only errors

alexandru-malaescu:
On expressions compilation failure, as part of the thrown exception, we were also adding the warnings resulting from the compilation process. This behavior is not consistent with the generated output of the equivalent functionality in Net Framework, which uses just the errors, without the warnings.

[UI-51453](https://uipath.atlassian.net/browse/UI-51453). Compilation warnings for missing namespace sources are visible as part of the error messages of all invalid activities for NET5 projects

### DIFF
--- a/src/UiPath.Workflow/Activities/ScriptingJitCompiler.cs
+++ b/src/UiPath.Workflow/Activities/ScriptingJitCompiler.cs
@@ -3,6 +3,7 @@
 
 using System.Activities.ExpressionParser;
 using System.Activities.Internals;
+using System.Activities.XamlIntegration;
 using System.Collections.Generic;
 using System.Linq;
 using System.Linq.Expressions;
@@ -72,8 +73,13 @@ public abstract class ScriptingJitCompiler : JustInTimeCompiler
         var results = ScriptingAotCompiler.BuildAssembly(finalCompilation);
         if (results.HasErrors)
         {
+            var errorResults = new TextExpressionCompilerResults
+            {
+                ResultType = results.ResultType,
+            };
+            errorResults.AddMessages(results.CompilerMessages.Where(m => !m.IsWarning));
             throw FxTrace.Exception.AsError(new SourceExpressionException(
-                SR.CompilerErrorSpecificExpression(expressionToCompile.Code, results), results.CompilerMessages));
+                SR.CompilerErrorSpecificExpression(expressionToCompile.Code, errorResults), errorResults.CompilerMessages));
         }
 
         return (LambdaExpression) results.ResultType.GetMethod("CreateExpression")!.Invoke(null, null);


src/UiPath.Workflow/Activities/ScriptingJitCompiler.cs
```diff
@@ -72,8 +73,13 @@ public override LambdaExpression CompileExpression(ExpressionToCompile expressio
         var results = ScriptingAotCompiler.BuildAssembly(finalCompilation);
         if (results.HasErrors)
         {
+            var errorResults = new TextExpressionCompilerResults
+            {
+                ResultType = results.ResultType,
+            };
+            errorResults.AddMessages(results.CompilerMessages.Where(m => !m.IsWarning));
```
dmetzgar, line 80: After filtering out the warnings you may have nothing left and shouldn't throw the exception.

dmetzgar:
@alexandru-malaescu I'm fine with this change but want to make sure we don't throw an exception if there are no error messages after filtering out warnings. VBV doesn't preserve that flag because the validation messages are passed via exception ([link](https://github.com/UiPath/CoreWF/blob/df30884bf75543a80c2fccdcd8d9a563c1a59819/src/UiPath.Workflow/Microsoft/VisualBasic/Activities/VisualBasicValue.cs#L125-L125)) so I think it's OK to ignore the warnings.

It is possible to get more granular control of the validation. If you set [ValidationSettings.ForceExpressionCache](https://github.com/UiPath/CoreWF/blob/df30884bf75543a80c2fccdcd8d9a563c1a59819/src/UiPath.Workflow.Runtime/Validation/ValidationSettings.cs#L83-L83) to false, a different code path is used to validate the expressions. This will no longer produce the LambdaExpression in VBV or CSV activities but will give you the ValidationError object with the IsWarning flag preserved. In addition, you can modify/filter/insert validation messages by modifying [ProcessDiagnostics](https://github.com/UiPath/CoreWF/blob/df30884bf75543a80c2fccdcd8d9a563c1a59819/src/UiPath.Workflow/Activities/RoslynExpressionValidator.cs#L126-L126). It's also [faster](https://github.com/UiPath/CoreWF/pull/178). But you can only use it if you don't need the LambdaExpression.

src/UiPath.Workflow/Activities/ScriptingJitCompiler.cs
alexandru-malaescu: Good point, although if we're inside the `results.HasErrors` if-block, we're already sure that we have at least 1 error message (which is not a warning). So filtering-out the warnings will never result in an empty list :)

dmetzgar (approved):

PR #213, opened 2022-08-29
fix 17 variables expression

mirceaarhip:


src/UiPath.Workflow/Activities/ScriptingJitCompiler.cs
```diff
@@ -116,16 +118,37 @@ public class VbJitCompiler : ScriptingJitCompiler
...
+        var part1 = new StringBuilder();
+        var part2 = new StringBuilder();
+
+        for (var i = 0; i < arrayType.Length - 1; i++)
+        {
+            part1.Append($" In T{i},");
+            part2.Append($" ByVal arg as T{i},");
+        }
+        part2.Remove(part2.Length - 1, 1);
+        return $"Public Delegate Function Func(Of {part1} Out TResult)({part2}) As TResult";
```
dmetzgar: @mirceaarhip @cosminsandu25 When the workflow is compiled all of the expressions are put into a single class. If there are two expressions that take more than 16 arguments and the arguments have the same types, I think there will be duplicate delegates. Is it possible to give the delegates unique names?

dmetzgar (approved):

PR #217, opened 2022-09-27
Remove lock around compiler usage

marius-bughiu:
This `lock` can introduce significant delay in certain scenarios when the validation is executed on multiple threads.
My test scenario:
- project with 24 files
- Studio attempts to validate using 8 threads
-> referenced assemblies for the files are rather similar resulting in only 4 cached compilers being created.
-> threads are fighting (locking) over these compilers; the more in-demand a certain compiler is, the more visible the effect 

In some test runs I did, we spent up to half the validation time just waiting for the compiler.

I've checked the underlying classes and there doesn't seem to be any state to warrant the lock. I mean, the only state would be the `MetadataReference` which we never modify. I've changed it to `IReadOnlyCollection` and replaced the `set` with an `init` to make this more obvious.

* As I understand it, this impacts runtime/execution as well (not just the validation flow).


dmetzgar (approved):
Thanks!

PR #224, opened 2022-12-06
Make light validation flow thread-safe, fix compilation

marius-bughiu:
There's a few takeways in this PR:
- removed `DesktopAssemblyIdentityComparer` from the compiler options as this was breaking compilation when the same assembly was provided twice but with different versions. This works in the current validation flow, so I expect the light validation should behave the same way.
- removed `OptionStrict.On` - it was causing all kinds of warnings in which we are not interested
- made `RoslynExpressionValidator` thread-safe. Studio Desktop will attempt to validate multiple files in parallel, so this needs to work
- added `_defaultNamespaces` in the `RoslynExpressionValidator` and included them part of the compilation. Without these namespaces, the compilation would fail with unknown type `Expression` errors. Since the `Expression` type is introduced by the validator for the purpose of validation/compilation, it makes sense for it to provide the required namespace.

In terms of testing, I've tested this by hand with the latest version of Studio with a couple larger projects I had at hand.

### DIFF
--- a/src/UiPath.Workflow/Activities/RoslynExpressionValidator.cs
+++ b/src/UiPath.Workflow/Activities/RoslynExpressionValidator.cs
@@ -27,6 +27,7 @@ public abstract class RoslynExpressionValidator
 
     private const string Comma = ", ";
     private readonly Lazy<ConcurrentDictionary<Assembly, MetadataReference>> _metadataReferences;
+    private readonly object _lockRequiredAssemblies = new();
 
     /// <summary>
     ///     Initializes the MetadataReference collection.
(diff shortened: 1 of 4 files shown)


src/UiPath.Workflow/Activities/RoslynExpressionValidator.cs
```diff
@@ -31,62 +37,35 @@ public abstract class RoslynExpressionValidator
...
 
-    /// <summary>
-    ///     Adds an assembly to the <see cref="RequiredAssemblies"/> set.
-    /// </summary>
-    /// <param name="assembly">assembly</param>
-    /// <remarks>
-    ///     Takes a lock and replaces <see cref="RequiredAssemblies"/> with a new set. Lock is taken in case
-    ///     multiple threads are adding assemblies simultaneously.
-    /// </remarks>
-    public void AddRequiredAssembly(Assembly assembly)
```
dmetzgar: This is used in Studio Web and it's public surface area. Is there a reason to remove it entirely?

marius-bughiu: That's my bad, I didn't realize it's part of the public API. I've restored it.

marius-bughiu:
@dmetzgar, - @aoltean16 did some further investigation on this and found another scenario where the validation wasn't working due to missing default namespaces. I've merged his changes in this branch.

src/UiPath.Workflow/Activities/VbExpressionValidator.cs
```diff
@@ -74,7 +74,6 @@ protected override void UpdateCompilationUnit(ExpressionContainer expressionCont
                 mainTypeName: null,
                 globalImports: globalImports,
                 rootNamespace: "",
-                optionStrict: OptionStrict.On,
```
dmetzgar, line 77: This will have some downstream effects so I want to be sure it's needed. My understanding is that at runtime strict is off. At design time Studio Web depends on strict to be on and goes through some extra steps to make sure the types are correct before validation. We can change the option in the designer but it's just important to make sure that happens. I'm not sure what Studio needs for strict. Not for this PR but I'm wondering if we should add an option somewhere to control strict without having to subclass the whole validator.

dmetzgar (approved):

dmetzgar:
There's a bunch of test failures for CSharpShould_Infer_Type and AheadOfTimeExpressions.CSharpCalculation where it seems unable to get the correct type. Could you take a look at those before merging?

marius-bughiu:
@aoltean16, looks like the tests fail due to changes in the last commit. Could you check them please? I'm not sure exactly which scenario you were covering with it, so I think it's better if you make the fix & re-test.

src/UiPath.Workflow/Activities/VbExpressionValidator.cs
marius-bughiu: The reason I disabled it is because it was causing warnings such as this one: 

```warning BC40057: Namespace or type specified in the project-level Imports 'System.Activities.DynamicUpdate' doesn't contain any public member or cannot be found. Make sure the namespace or the type is defined and contains at least one public member. Make sure the imported element name doesn't use any aliases.	```

-- one for each expression it compiles, so you'd end up with a lot of them in any decent project.

image

From what I tested, this didn't happen on the old validation flow, so I removed the option in order to maintain the existing behavior. If you have any other idea as to what might be causing this instead, I could look more into it.

aoltean16:
all good now, it was an oversight from my side with the included default namespaces.

aoltean16:
@dmetzgar can you please merge this?

src/UiPath.Workflow/Activities/VbExpressionValidator.cs
dmetzgar: VB is annoying. There's no way to get rid of these warnings without turning off strict.

PR #227, opened 2023-01-10
Clean and adapt for light validation use [STUD-61340]

aoltean16:
`VBValue` and `VBReference` were compiling the expression and storing it as a lambda when calling `CacheMetadata`. While the reasons for this behavior might be forgotten, as long as `VbValue`, and `VbReference` (same as `CSharpValue` and `CSharpReference`) requires compilation, there is no reason to have this overhead in there.
The current implementation has a big performance impact in Studio, while running a workflow with 2000 expressions (which is quite common for clients). In this case, the RAM spikes up to more than 12gb, and it takes more than 2 minutes to validate the workflow, on a high-end machine (see https://uipath.atlassian.net/wiki/spaces/STUD/pages/87044198116/NetCore+validation+performance+improvements for a comparison).
Using the new validation, it takes roughly 5-7 seconds, and the RAM never spikes more than 600mb.

Moreover, since the `CSharp` and `VisualBasic` implementation both requires compilation now, there is no reason to have different implementations.

### DIFF
--- a/src/UiPath.Workflow/Activities/RoslynExpressionValidator.cs
+++ b/src/UiPath.Workflow/Activities/RoslynExpressionValidator.cs
@@ -177,7 +177,11 @@ public abstract class RoslynExpressionValidator
         PrepValidation(expressionContainer);
 
         ModifyPreppedCompilationUnit(expressionContainer);
-        var diagnostics = expressionContainer.CompilationUnit.GetDiagnostics().Select(diagnostic =>
+        var diagnostics = expressionContainer
+            .CompilationUnit
+            .GetDiagnostics()
+            .Where(d=> d.Severity == DiagnosticSeverity.Error)
+            .Select(diagnostic =>
             new TextExpressionCompilerError
             {
                 SourceLineNumber = diagnostic.Location.GetMappedLineSpan().StartLinePosition.Line,
--- a/src/UiPath.Workflow/Microsoft/VisualBasic/Activities/VisualBasicReference.cs
+++ b/src/UiPath.Workflow/Microsoft/VisualBasic/Activities/VisualBasicReference.cs
@@ -12,7 +12,6 @@ using System.ComponentModel;
 using System.Diagnostics;
 using System.Linq.Expressions;
 using System.Windows.Markup;
-using ActivityContext = System.Activities.ActivityContext;
 
 namespace Microsoft.VisualBasic.Activities;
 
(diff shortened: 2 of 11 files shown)


src/UiPath.Workflow/Activities/RoslynExpressionValidator.cs
```diff
@@ -177,7 +177,11 @@ protected virtual IEnumerable<ValidationError> ProcessDiagnostics(ExpressionCont
         PrepValidation(expressionContainer);
 
         ModifyPreppedCompilationUnit(expressionContainer);
-        var diagnostics = expressionContainer.CompilationUnit.GetDiagnostics().Select(diagnostic =>
+        var diagnostics = expressionContainer
+            .CompilationUnit
+            .GetDiagnostics()
+            .Where(d=> d.Severity == DiagnosticSeverity.Error)
```
dmetzgar, line 183: I take it for designer we don't care about warnings.

src/UiPath.Workflow/Microsoft/VisualBasic/Activities/VisualBasicReference.cs
```diff
@@ -12,17 +12,14 @@
...
 using System.Windows.Markup;
-using ActivityContext = System.Activities.ActivityContext;
 
 namespace Microsoft.VisualBasic.Activities;
 
 [DebuggerStepThrough]
 public sealed class VisualBasicReference<TResult> : CodeActivity<Location<TResult>>, IValueSerializableExpression,
     IExpressionContainer, ITextExpression
 {
-    private Expression<Func<ActivityContext, TResult>> _expressionTree;
```
dmetzgar, line 23: Removing the expression tree is very interesting. I had looked before to see if there were any dependencies on it and it seemed there were. But if this expression tree can be removed going forward I'm all for it.

dmetzgar (approved):

src/UiPath.Workflow/Activities/RoslynExpressionValidator.cs
aoltean16: we don't really :)

src/UiPath.Workflow/Microsoft/VisualBasic/Activities/VisualBasicReference.cs
aoltean16: considering we set RequiresCompilation = true for VB too, this does not make any sense now. The only difference is now you need to compile the workflow, before you can access the expression tree of an ITextExpression (which was already the case for C#). Currently we need to have SWeb on board, as they seem to be using GetExpressionTree.

PR #228, opened 2023-01-11
Support collectible assemblies for JIT compilation and fix memory leak

AlvinStanescu:
Given the need for supporting collectible assemblies loaded in our designer, we should also support JIT compiling these (given a loaded or loadable compiled assembly).

Additionally, the `Assembly.Load` on a temporary assembly was leaking memory (consider that this was done for every expression, in each workflow).

### DIFF
--- a/src/UiPath.Workflow/Activities/ScriptingJitCompiler.cs
+++ b/src/UiPath.Workflow/Activities/ScriptingJitCompiler.cs
@@ -75,7 +75,7 @@ public abstract class ScriptingJitCompiler : JustInTimeCompiler
         var finalCompilation = compilation.ReplaceSyntaxTree(syntaxTree, syntaxTree.WithChangedText(SourceText.From(
             CreateExpressionCode(types, names, expressionToCompile.Code))));
         
-        var collectibleAlc = new AssemblyLoadContext("ScriptingAot" + Guid.NewGuid(), true);
+        var collectibleAlc = new AssemblyLoadContext("ScriptingJit" + Guid.NewGuid(), true);
         collectibleAlc.Resolving += CollectibleAlc_Resolving;
         using var scope = collectibleAlc.EnterContextualReflection();
         
--- a/src/UiPath.Workflow/Activities/ScriptingJitCompiler.cs
+++ b/src/UiPath.Workflow/Activities/ScriptingJitCompiler.cs
@@ -9,6 +9,7 @@ using System.Linq;
 using System.Linq.Expressions;
 using System.Reflection;
 using System.Reflection.Metadata;
+using System.Runtime.Loader;
 using System.Text;
 using System.Threading;
 using Microsoft.CodeAnalysis;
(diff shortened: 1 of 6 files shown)


src/UiPath.Workflow/Activities/ScriptingJitCompiler.cs
```diff
@@ -73,7 +74,12 @@ public override LambdaExpression CompileExpression(ExpressionToCompile expressio
                 .Select(GetTypeName));
         var finalCompilation = compilation.ReplaceSyntaxTree(syntaxTree, syntaxTree.WithChangedText(SourceText.From(
             CreateExpressionCode(types, names, expressionToCompile.Code))));
-        var results = ScriptingAotCompiler.BuildAssembly(finalCompilation);
+        
+        var collectibleAlc = new AssemblyLoadContext("ScriptingAot" + Guid.NewGuid(), true);
```
AlvinStanescu: I can also hide this under an internal "feature flag" so we can only enable this in Studio (for validate in designers). Let me know if this is preferable. 

For AOT compilation, we still use the standard assembly load context.

src/UiPath.Workflow/Activities/ScriptingJitCompiler.cs
```diff
@@ -73,7 +74,12 @@ public override LambdaExpression CompileExpression(ExpressionToCompile expressio
                 .Select(GetTypeName));
         var finalCompilation = compilation.ReplaceSyntaxTree(syntaxTree, syntaxTree.WithChangedText(SourceText.From(
             CreateExpressionCode(types, names, expressionToCompile.Code))));
-        var results = ScriptingAotCompiler.BuildAssembly(finalCompilation);
+        
+        var collectibleAlc = new AssemblyLoadContext("ScriptingJit" + Guid.NewGuid(), true);
```
dmetzgar, line 78: Do you need to explicitly unload the AssemblyLoadContext?

src/UiPath.Workflow/Activities/ScriptingJitCompiler.cs
```diff
@@ -88,6 +94,15 @@ public override LambdaExpression CompileExpression(ExpressionToCompile expressio
         return (LambdaExpression)results.ResultType.GetMethod("CreateExpression")!.Invoke(null, null);
     }
 
+    private Assembly CollectibleAlc_Resolving(AssemblyLoadContext loadContext, AssemblyName assemblyName)
```
dmetzgar, line 97: This all looks good. I'm curious how the AssemblyLoadContext knows enough to have the full AssemblyName but can't load the Assembly on its own.

src/UiPath.Workflow/Activities/ScriptingJitCompiler.cs
AlvinStanescu: We don't need to, as it should get garbage collected away (along with its assemblies) once there are no strong references to any assembly loaded in that context.

However, there is a caveat here - some WPF services (see `SafeSecurityHelper.cs` in multiple WPF assemblies) hold on to these collectible assemblies. 

The logic is there for them to not hold on to dynamic assemblies, but not for collectible ones. Tempted to also make a PR on https://github.com/dotnet/wpf for this, but didn't have the time. 

In Studio we do some reflection magic to clear collectible assemblies from these Safe Security Helpers once in a while, so in practice this should be solved.

src/UiPath.Workflow/Activities/ScriptingJitCompiler.cs
AlvinStanescu: I must confess that it also seemed strange to me that even when I used a collectible AssemblyLoadContext, it was always searching for it using the full name at this stage, but for some reason this always lead to a standard `Assembly.Load` on that assembly (triggered by the `CreateExpression` call, as the expression used a collectible assembly value), which failed as this load context is collectible. 

Didn't really investigate this in-depth, but I suspect this may be a .NET 6 bug which may not repro in .NET 7 since I know that they rewrote some parts of the collectible assemblies handling.

dmetzgar (approved):

PR #232, opened 2023-02-06
Expression improvements

aoltean16:
brought back a bit of the old code that was deleted in my previous PR, and more.
Exposed a way to execute an ITextExpression at a given context. This was done in Studio for VB and was lacking proper support for C#.
Also made sure GetExpressionTree will now return back the original expression even if the workflow is not compiled. This should help SWeb and Studio with certain functionalities that was relying on it.

### DIFF
--- a/src/UiPath.Workflow/UiPath.Workflow.csproj
+++ b/src/UiPath.Workflow/UiPath.Workflow.csproj
@@ -30,6 +30,9 @@
         <ItemGroup>
             <BuildOutputInPackage Include="$(OutputPath)Microsoft.CodeAnalysis.VisualBasic.Scripting.*"/>
         </ItemGroup>
+		<ItemGroup Condition="'$(TargetFramework)' == 'net6.0' AND '$(Configuration)'=='Debug'">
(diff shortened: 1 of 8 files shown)


src/UiPath.Workflow/UiPath.Workflow.csproj
```diff
@@ -30,6 +30,9 @@
         <ItemGroup>
             <BuildOutputInPackage Include="$(OutputPath)Microsoft.CodeAnalysis.VisualBasic.Scripting.*"/>
         </ItemGroup>
+		<ItemGroup Condition="'$(TargetFramework)' == 'net6.0' AND '$(Configuration)'=='Debug'">
```
dmetzgar, line 33: Was this missing before?

dmetzgar (approved):

src/UiPath.Workflow/UiPath.Workflow.csproj
aoltean16: helps debugging with Studio :). No real use in Release mode

PR #237, opened 2023-02-14
Include fallback if an expression is not compiled

aoltean16:
If an expression is not compiled, compile it in Execute method and run it. This is so we do not break old workflows, or non-compliant activities that use VisualBasicValue / Reference in places other than on public, browsable properties of an activity.


marius-bughiu (approved):
Haven't tested it by hand, relying on the unit tests to hold water.

PR #243, opened 2023-02-21
Let ScriptingJitCompiler resolve assemblies, which already exist in AppDomain

GerardCieciwa:
Sometimes there is a need to compile expressions, which require references to assemblies, which do exist in AppDomain.CurrentDomain, but not in current AssemblyLoadContext. In such a case current implementation is looking for such an assembly in AppDomain.CurrentDomain, but doesn't load it. This causes exceptions when last line of ScriptingJitCompiler.CompileExpression ("GetMethod("CreateExpression")!.Invoke") is called. 
According to https://learn.microsoft.com/en-us/dotnet/api/System.Runtime.Loader.AssemblyLoadContext?view=net-6.0 (section "Application usage") AssemblyLoadContext.Resolving is a "second chance to resolve the assembly" after call to  AssemblyLoadContext.Load, so it's logical, it should do the same (i.e. not  only resolve, but also load). Current implementation resolves and loads when assembly isn't found in AppDomain.CurrentDomain, but only resolves otherwise. 
This fix makes attempt to load always.

### DIFF
--- a/src/UiPath.Workflow/Activities/ScriptingJitCompiler.cs
+++ b/src/UiPath.Workflow/Activities/ScriptingJitCompiler.cs
@@ -97,9 +97,8 @@ public abstract class ScriptingJitCompiler : JustInTimeCompiler
     private Assembly CollectibleAlc_Resolving(AssemblyLoadContext loadContext, AssemblyName assemblyName)
     {
         var assembly = AppDomain.CurrentDomain.GetAssemblies().FirstOrDefault(a => a.FullName == assemblyName.FullName);
-        if (assembly != null)
-            return assembly;
-
+        if (assembly is not null && !string.IsNullOrWhiteSpace(assembly.Location))
+            return loadContext.LoadFromAssemblyPath(assembly.Location);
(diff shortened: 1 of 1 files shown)


dmetzgar (approved):

AlvinStanescu (changes requested):
I don't think the fix is the correct one for this scenario. I think we simply have another assembly already loaded and (luckily) we're able to load an updated one from disk and this works, but this is just a coincidence.

If you look at https://github.com/UiPath/CoreWF/pull/228, which introduced the load context over here, it was merged after Jan. 04, which was the date this bug was filed. So fixing this here is imo not a good idea as we're only fixing a symptom.

Instead, we should try to understand why the entities assembly that is loaded is not the latest one (the one created after we updated the field). Or, if two assemblies of the same name are loaded, at least we should try to load the newest one from the current app domain, but without re-loading it for each expression.

marius-bughiu:
The issue which generated this PR was found to be in Studio and fixed there.

PR #249, opened 2023-03-06
Remove unnecessary code from activity visitors

aoltean16:
No behavior changed, simply removed unnecessary argument from a lot of methods.


dmetzgar (approved):
I'm wondering if it was there to support dynamic update. We have no plans to implement dynamic update though.

PR #252, opened 2023-03-14
Let ScriptingJitCompiler resolve assemblies, which are passed as references

GerardCieciwa:
This is improved version of https://github.com/UiPath/CoreWF/pull/243
In this version;
1. Resolved references are sought among assemblies passed as referencedAssemblies. This should lead to consistent assembly resolution between compilation and compiled method invocation (`results.ResultType.GetMethod("CreateExpression")!.Invoke(null, null)`)
2. AppDomain can contain (possibly obsolete) assemblies, which were loaded to AssemblyLoadContext which is already unloaded, what could lead to unexpected results. This version narrows the scope of search to assemblies explicitly designated as  references.
3. Explicit unload of temporary AssemblyLoadContext was added.


marius-bughiu:
The issue which generated this PR was found to be in Studio and fixed there.

PR #254, opened 2023-03-24
Allow compilation/validation caches to be cleared

AlvinStanescu:


vuplea:
As discussed with @lbargaoanu, changes are OK w.r.t. runtime (robot) behavior.

aoltean16 (approved):

AlvinStanescu:
Fix re-validated on latest Studio 23.4 beta build and assembly collection is performed as expected, for both C# and VB validation.

PR #255, opened 2023-04-04
Fix vb issue with identifiers uniqueness

aoltean16:
The issue comes when you have a variable with the same name as a method, i.e. variable name: count, of type `List<Object>` and writing the following expression: `count.Count.ToString`. Aligned the behavior to be the same as ScriptingJitCompiler, so VBValidator will treat identifiers unique ignoring case. CSValidator works as expected.


lbargaoanu (approved):
Some code reuse is in order.

PR #260, opened 2023-05-23
Improve first invoke performance

angelowang:
The first `Invoke()` call is slow. When the xaml file contains a lot of VB scripts, JIT compiler will be used, but for **every** compilation, the compiler starts from the scratch and always tries to resolve missing references via `MetadataReference.CreateFromFile`, thus it will be very slow. With some test cache between compilations, one of our test xaml file is improved from 18s to 3.5s.

### DIFF
--- a/src/UiPath.Workflow/Activities/ScriptingJitCompiler.cs
+++ b/src/UiPath.Workflow/Activities/ScriptingJitCompiler.cs
@@ -4,11 +4,14 @@
 using System.Activities.ExpressionParser;
 using System.Activities.Internals;
 using System.Activities.XamlIntegration;
+using System.Collections.Concurrent;
 using System.Collections.Generic;
+using System.Collections.Immutable;
 using System.Linq;
 using System.Linq.Expressions;
 using System.Reflection;
 using System.Reflection.Metadata;
+using System.Runtime.Collections;
 using System.Runtime.Loader;
 using System.Text;
 using System.Threading;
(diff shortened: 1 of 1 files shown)


src/UiPath.Workflow/Activities/ScriptingJitCompiler.cs
```diff
@@ -36,6 +39,133 @@ public record ExpressionToCompile(string Code, IReadOnlyCollection<string> Impor
     : CompilerInput(Code, ImportedNamespaces)
 { }
 
+public sealed class CachedMetadataReferenceResolver : MetadataReferenceResolver
+{
+    public static CachedMetadataReferenceResolver Default = new CachedMetadataReferenceResolver(ScriptMetadataResolver.Default);
+
+    ScriptMetadataResolver _resolver;
+
+    private class ResolveCacheKey
```
dmetzgar, line 48: Can you use a record for this instead? It already has equals and hashcode implementations that do very similar things to what you're doing here.

src/UiPath.Workflow/Activities/ScriptingJitCompiler.cs
```diff
@@ -36,6 +39,133 @@ public record ExpressionToCompile(string Code, IReadOnlyCollection<string> Impor
...
+            return _reference == rtcKey._reference &&
+                _baseFilePath == rtcKey._baseFilePath &&
+                _properties.Equals(rtcKey._properties);
+        }
+
+        public override int GetHashCode() => _hashCode;
+    }
+    ConcurrentDictionary<ResolveCacheKey, ImmutableArray<PortableExecutableReference>> _resolveCache = new ConcurrentDictionary<ResolveCacheKey, ImmutableArray<PortableExecutableReference>>();
+
+    private class ResolveMissingCacheKey
```
dmetzgar, line 83: This could also be a record.

src/UiPath.Workflow/Activities/ScriptingJitCompiler.cs
```diff
@@ -36,6 +39,133 @@ public record ExpressionToCompile(string Code, IReadOnlyCollection<string> Impor
...
+        public override int GetHashCode() => _hashCode;
+    }
+    ConcurrentDictionary<ResolveMissingCacheKey, PortableExecutableReference> _resolveMissingCache = new ConcurrentDictionary<ResolveMissingCacheKey, PortableExecutableReference>();
+
+    public CachedMetadataReferenceResolver(ScriptMetadataResolver resolver)
+    {
+        _resolver = resolver;
+    }
+
+    public override bool Equals(object other)
```
dmetzgar, line 119: Why does the resolver need this if you're always using the singleton?

dmetzgar (changes requested):
Thanks for submitting this PR and apologies for the delay in reviewing. If you still want to go forward, this code seems it could be a lot simpler using records. There are enough custom hash code implementations in WF and I'd like to avoid adding another one.

aoltean16 (changes requested):
Please provide an unit test or benchmark showcasing the issue and the improvement. 
Thank you!

PR #268, opened 2023-06-14
Compile all expressions at once when validating.

aoltean16:


angelowang:
Hi @aoltean16 , will this change improve the performance? I had to make some change like in https://github.com/UiPath/CoreWF/pull/260 to make it much faster on first Run(). There are many duplicated file access which is time consuming when we have a lot of CodeActivities. But even after that, the performance is still 10x slower than the VbHostedCompiler before :(

Seems it's only for design time?

src/UiPath.Workflow.Runtime/Validation/ActivityValidationServices.cs
```diff
@@ -453,9 +454,28 @@ internal ValidationResults InternalValidate()
                 ActivityUtilities.CacheRootMetadata(_rootToValidate, _environment, _options, new ActivityUtilities.ProcessActivityCallback(ValidateElement), ref _errors);
             }
 
+            var extension = GetValidationExtension(_environment);
```
vuplea: As expressions are the concern of UiPath.Workflow project, expression validation should also be implemented there.

aoltean16:
> Hi @aoltean16 , will this change improve the performance? I had to make some change like in #260 to make it much faster on first Run(). There are many duplicated file access which is time consuming when we have a lot of CodeActivities. But even after that, the performance is still 10x slower than the VbHostedCompiler before :(
> 
> Seems it's only for design time?

This improves the validation time drastically, when called by ActivityValidationServices.Validate() with ValidationSettings.ForceExpressionsCache = false, so in theory, yea, only design time will benefit from this

angelowang:
>in theory, yea, only design time will benefit from this

Thanks @aoltean16 . The execution time also needs this eagerly. We have a case where 21 xaml files takes 30s for the first time running. This is still the time after the MetadataReference caching, the original time could be well over several minutes.

The idea might be similar, like each expression's LambdaExpression could be from a separate `CreateExpression_actionHash/codeHash/actionId` function.

src/Test/TestCases.Workflows/ExpressionTests.cs
```diff
@@ -424,6 +426,67 @@ public void CSRoslynValidator_ValidatesMoreThan16Arguments()
...
+                    To = new OutArgument<int>(new CSharpReference<int>("sum")),
+                    Value = new InArgument<int>(new CSharpValue<int>($"sum + {j}"))
+                });
+            }
+        }
+        foreach (var activity in activities)
+        {
+            tasks.Add(Task.Run(() =>
+            {
+                results.Add(ActivityValidationServices.Validate(activity, _useValidator));
```
gabriela-lungu-uip, line 485: as foreach iterates on an empty list, validate is never called

vuplea (approved):

gabriela-lungu-uip (approved):

PR #275, opened 2023-07-19
fix: sw compatibility

liviu-uba:
- unseal ExpressionValidator classes
- allow curation of validation errors produces by compilation


aoltean16 (approved):
LGTM

PR #282, opened 2023-09-12
Fixed compatibility issue between 23.4 and 23.10 by removing the base class TextExpressionBase

aoltean16:


src/UiPath.Workflow/Microsoft/VisualBasic/Activities/TextExpressionHelper.cs
```diff
@@ -1,20 +1,12 @@
...
         private static readonly Func<ValidationExtension> _validationFunc = () => new();
 
-        public abstract string ExpressionText { get; set; }
-
-        public abstract string Language { get; }
-
-        public abstract Expression GetExpressionTree();
-
-        protected bool QueueForValidation<T>(CodeActivityMetadata metadata, bool isLocation)
+        public static bool QueueForValidation<T>(Activity activity, CodeActivityMetadata metadata, string expressionText, string language, bool isLocation)
```
david-barzi, line 9: Can't this be simplified by replacing `Activity` with `ITextExpression` (or casting it) and getting the expression text and language directly, instead of passing them?

aoltean16: Can't replace `Activity` with `ITextExpression`, because it needs activity info further down the line, and combined with `ITextExpression` fields(as you mentioned, `ExpressionText `and `Language`).  In my opinion, casting it to `ITextExpression` would simplify the call, but would look a bit weird to do so in the implemented method (this case we do know for sure that activity is `ITextExpression`, but if it isnt? dont want to shortcut it to return false, as it would beat the purpose of this alltogether).

david-barzi (approved):

mirceaarhip (approved):

PR #286, opened 2023-11-07
Compile expressions but do not emit the assembly.

aoltean16:
The benefits are documented here
https://uipath.atlassian.net/wiki/spaces/STUD/pages/87410606685/ExpressionTextbox+typing+improvement


src/Perf/CoreWf.Benchmarks/ExpressionTextboxPerformance.cs
```diff
@@ -0,0 +1,26 @@
+﻿//using BenchmarkDotNet.Attributes;
+//using Microsoft.CSharp.Activities;
+//using Microsoft.VisualBasic.Activities;
+//using System.Activities;
+
+//namespace CoreWf.Benchmarks
+//{
+//    [MemoryDiagnoser(true)]
+//    public class ExpressionTextboxPerformance
```
mirceaarhip: why is this code commented ?

src/UiPath.Workflow/Activities/DesignerHelperImpl.cs
```diff
@@ -75,106 +87,42 @@ internal abstract class DesignerHelperImpl
...
-            {
-                if (ns != null)
-                {
-                    namespacesSet.Add(ns);
-                }
-            }
-        }
+        // execute compiler
+        var compilation = Compiler.Compile(expressionText, isLocation, targetType ?? typeof(object), namespaces.ToList(), referencedAssemblies.ToList(), environment);
+        var diagnostics = Task.Run(async () => await compilation.WithAnalyzers(_usedTypesAnalizerList).GetAllDiagnosticsAsync()).Result;
```
mirceaarhip: blazor is mono thread so .Result might block our precious thread

src/UiPath.Workflow/Activities/DesignerHelperImpl.cs
```diff
@@ -75,106 +87,42 @@ internal abstract class DesignerHelperImpl
...
-            foreach (var ns in namespaces)
-            {
-                if (ns != null)
-                {
-                    namespacesSet.Add(ns);
-                }
-            }
-        }
+        // execute compiler
+        var compilation = Compiler.Compile(expressionText, isLocation, targetType ?? typeof(object), namespaces.ToList(), referencedAssemblies.ToList(), environment);
```
mirceaarhip, line 409: previous code ensures namespace/ referenced assembly are not null

src/UiPath.Workflow/Activities/DesignerHelperImpl.cs
```diff
@@ -414,23 +227,23 @@ private static void FindTypeReferences(Expression expression, HashSet<Type> type
...
             case ExpressionType.Subtract:
             case ExpressionType.SubtractChecked:
-                var binaryExpression = (BinaryExpression) expression;
+                var binaryExpression = (BinaryExpression)expression;
                 FindTypeReferences(binaryExpression.Left, typeReferences);
                 FindTypeReferences(binaryExpression.Right, typeReferences);
                 return;
 
             case ExpressionType.Conditional:
-                var conditional = (ConditionalExpression) expression;
```
mirceaarhip, line 423: nit: i would recommend to split formatting changes to another PR , every reviewer will appreciate

src/UiPath.Workflow/Activities/DesignerHelperImpl.cs
```diff
@@ -713,4 +526,29 @@ private static void FindTypeReferences(MemberBinding binding, HashSet<Type> type
...
+            }));
+        return new SourceExpressionException(
+            SR.CompilerErrorSpecificExpression(expressionText, errors), errors.CompilerMessages);
+    }
+
+    private static void PopulateVbSettings(VisualBasicSettings vbSettings, ImmutableArray<Diagnostic> diagnostics)
+    {
+        foreach (var diagnostic in diagnostics.Where(d => d.Id == _usedTypesDiagnosticId))
+        {
+            var diagMessages = diagnostic.GetMessage()[1..^1].Split('|');
```
mirceaarhip, line 836: let's make sure that all this string manipulations are covered by tests and the cannot failed with null reference

src/UiPath.Workflow/Activities/ScriptingJitCompiler.cs
```diff
@@ -43,7 +43,7 @@ protected ScriptingJitCompiler(HashSet<Assembly> referencedAssemblies)
     protected IReadOnlyCollection<MetadataReference> MetadataReferences { get; init; }
     protected abstract string GetTypeName(Type type);
     protected abstract Script<object> Create(string code, ScriptOptions options);
-    protected abstract CompilerHelper CompilerHelper { get; }
```
mirceaarhip: this is a breaking change; sw actually uses this member

src/UiPath.Workflow/Activities/Utils/CompilerHelper.cs
```diff
@@ -1,15 +1,31 @@
-﻿using System;
+﻿using Microsoft.CodeAnalysis;
+using System.Collections.Generic;
 
 namespace System.Activities
 {
-    public abstract class CompilerHelper
+    internal abstract class CompilerHelper
```
mirceaarhip: breaking change

src/UiPath.Workflow/Microsoft/CSharp/Activities/CSharpDesignerHelper.cs
```diff
@@ -31,13 +31,9 @@ internal class CSharpExpressionFactory<T> : ExpressionFactory
 internal class CSharpDesignerHelperImpl : DesignerHelperImpl
 {
     public override Type ExpressionFactoryType => typeof(CSharpExpressionFactory<>);
-    public override string Language => "C#";
+    public override string Language => CSharpHelper.Language;
 
-    public override JitCompilerHelper CreateJitCompilerHelper(string expressionText, HashSet<AssemblyName> references,
```
mirceaarhip, line 36: i think this allows injection of a compiler something which sw does (not sure if it good or not) but with this changes we won't be able to to this

mirceaarhip (changes requested):

src/UiPath.Workflow/Activities/DesignerHelperImpl.cs
aoltean16: They are covered indirectly in tests, but I've added 2 more just for sanity, that directly checks import references, and will make it fail if any string manipulation fails (tho, we own that diagnostic so unless we change it, it wont fail).

src/UiPath.Workflow/Roslyn/UsedTypesAnalyzer.cs
```diff
@@ -0,0 +1,208 @@
...
+{
+#pragma warning disable RS2008
+    private static readonly DiagnosticDescriptor _usedTypesRule = new DiagnosticDescriptor("UT_001", "Used reference", "'{0}'", "Architecture", DiagnosticSeverity.Info, true);
+#pragma warning restore RS2008
+
+    public override ImmutableArray<DiagnosticDescriptor> SupportedDiagnostics { get; } = ImmutableArray.Create(_usedTypesRule);
+
+    public override void Initialize(AnalysisContext context)
+    {
+        context.EnableConcurrentExecution();
```
marius-bughiu, line 24: We might need to add a check here for SW, similar to the one found in the compiler helpers for the `concurrentBuild` - I mean, I assume this is what this check does? 
 `!RuntimeInformation.IsOSPlatform(OSPlatform.Create("BROWSER"))`

src/UiPath.Workflow/Roslyn/UsedTypesAnalyzer.cs
```diff
@@ -0,0 +1,208 @@
...
+    }
+
+    private sealed class CSharpReferenceCounter : CSharpSyntaxWalker, ITypeIdentifier
+    {
+        private readonly HashSet<string> _usedTypes;
+        private readonly SemanticModel _semanticModel;
+        private readonly SyntaxTree _syntaxTree;
+        private readonly CancellationToken _cancellationToken;
+
+        public CSharpReferenceCounter(SyntaxTree syntaxTree, SemanticModel semanticModel, HashSet<string> usedTypes, CancellationToken cancellationToken)
```
marius-bughiu: It is a bit weird to pass the `cancellationToken` in the constructor as it effectively renders the instance single-use. We should move it as a parameter to the methods.

src/UiPath.Workflow/Roslyn/UsedTypesAnalyzer.cs
```diff
@@ -0,0 +1,208 @@
...
+            ? new CSharpReferenceCounter(syntaxTree, semanticModel, usedTypes, context.CancellationToken)
+            : new VisualBasicReferenceCounter(syntaxTree, semanticModel, usedTypes, context.CancellationToken);
+    }
+
+    private interface ITypeIdentifier
+    {
+        void FindUsedTypes();
+    }
+
+    private sealed class CSharpReferenceCounter : CSharpSyntaxWalker, ITypeIdentifier
```
marius-bughiu: nit: the class name is a tiny bit confusing, since we don't really count references, we find used types. Maybe something along the lines `CSharpTypeIdentifier` would be more appropriate - would match the interface name as well (`ITypeIdentifier`).

src/UiPath.Workflow/Activities/Utils/CSharpCompilerHelper.cs
```diff
@@ -7,14 +12,34 @@ namespace System.Activities
...
+        private static readonly dynamic s_typeNameFormatter = GetTypeNameFormatter();
+        private static readonly dynamic s_typeOptions = GetTypeOptions();
+
+        public override Compilation DefaultCompilationUnit { get; } = InitDefaultCompilationUnit();
 
         public override int IdentifierKind => (int)SyntaxKind.IdentifierName;
 
-        public override StringComparer IdentifierNameComparer => StringComparer.Ordinal;
+
+        public override ParseOptions ScriptParseOptions { get; } = new CSharpParseOptions(kind: SourceCodeKind.Script);
```
marius-bughiu: You could in theory get rid of all the `as CSharpParseOptions` and `as CSharpCompilerHelper`, etc. by specifying the actual type in the override, like in this example you could do:

```
public override CSharpParseOptions ScriptParseOptions { get; } = new CSharpParseOptions(kind: SourceCodeKind.Script);
```

src/UiPath.Workflow/Activities/Utils/CSharpCompilerHelper.cs
```diff
@@ -7,14 +12,34 @@ namespace System.Activities
...
+
+        public override StringComparer IdentifierNameComparer { get; } = StringComparer.Ordinal;
+
+        public override string GetTypeName(Type type) =>
+            (string)s_typeNameFormatter.FormatTypeName(type, s_typeOptions);
+
+        private static object GetTypeNameFormatter()
```
marius-bughiu: nit: Move private method below the public ones.

src/UiPath.Workflow/Roslyn/UsedTypesAnalyzer.cs
```diff
@@ -0,0 +1,208 @@
...
+
+        foreach (var type in usedTypes)
+        {
+            context.ReportDiagnostic(Diagnostic.Create(_usedTypesRule, null, type));
+        }
+    }
+
+    private static ITypeIdentifier GetReferenceCounter(in CompilationAnalysisContext context, SyntaxTree syntaxTree, SemanticModel semanticModel, HashSet<string> usedTypes)
+    {
+        return context.Compilation.Language == "C#"
```
marius-bughiu: nit: `CSharpHelper.Language` instead of `"C#"`?

src/UiPath.Workflow/Activities/ExpressionCompiler.cs
```diff
@@ -0,0 +1,79 @@
...
+        }
+
+        protected static Type GetSystemType(ITypeSymbol typeSymbol, Assembly assembly)
+        {
+            StringBuilder typeName = new();
+            typeName.Append(GetTypeName(typeSymbol, assembly));
+            return assembly.GetType(typeName.ToString());
+        }
+
+        private static string GetTypeName(ITypeSymbol typeSymbol, Assembly assembly)
```
marius-bughiu: The `assembly` parameter is passed around but is not used for anything.

src/UiPath.Workflow/Activities/Utils/CSharpCompilerHelper.cs
aoltean16: since when is this a thing? damn life changer!

src/UiPath.Workflow/Roslyn/UsedTypesAnalyzer.cs
aoltean16: it adds unnecessary overhead for us and I dont think this is really a problem for them.

src/UiPath.Workflow/Roslyn/UsedTypesAnalyzer.cs
aoltean16: removed it from Visit method altogether, as the whole operation is under 1ms, the overhead of checking is not worth it anyway.

marius-bughiu (approved):

mirceaarhip (approved):

PR #290, opened 2023-11-16
Expose UnaryExpressionHelpers and BinaryExpressionHelpers

### LINKED ISSUE #288: Expose *ExpressionHelpers
ExpressionHelpers such as [BinaryExpressionHelper](https://github.com/UiPath/CoreWF/blob/develop/src/UiPath.Workflow.Runtime/Expressions/BinaryExpressionHelper.cs) are useful for developers of third party libraries, such as when attempting to implement #283 outside of CoreWF.

Please expose UnaryExpressionHelper and BinaryExpressionHelper.

TheBrambleShark:
Also promotes my Increment/Decrement and AddAssign/SubtractAssign expressions.

My added expressions are fully documented and ready for NET7. IfDefs can be removed if this is not desired.

Code also assumes a nullable context. I invite discussion on whether to remove nullable annotations or whether to enable nullable. I personally find it is more clear to the developer but this can be removed if requested.

All of these expressions expect numeric types (with appropriate constraints in Net7). In net6, theoretically any type would be allowed, but non-numeric types will throw. This is by design. The expressions simply provide activities for common C# operations.

`Increment` = `number++`
`Decrement` = `number--`
`AddAssign` = `number += otherNumber`
`SubtractAssign` = `number -= otherNumber`

If anyone knows how to add the designer files for these activities, please do let me know. I have been unable to figure out how to do this in Net6 and newer as PresentationCore does not seem to have been ported.

Edit: Fixes #288

### DIFF
--- a/src/UiPath.Workflow.Runtime/Expressions/AddAssign.cs
+++ b/src/UiPath.Workflow.Runtime/Expressions/AddAssign.cs
@@ -1,0 +1,98 @@
+﻿// This file is part of Core WF which is licensed under the MIT license.
+// See LICENSE file in the project root for full license information.
+
+using System.Activities.Validation;
+using System.ComponentModel;
+using System.Linq.Expressions;
+using System.Numerics;
+
+namespace System.Activities.Expressions
+{
+    /// <summary>
+    /// A code activity which incrementally assigns a numeral.
+    /// </summary>
+    /// <example>
+    /// int myNum = 5;
+    /// myNum += 5; // 10.
+    /// </example>
+    /// <typeparam name="TLeft">The operand to modify.</typeparam>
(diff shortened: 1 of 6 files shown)


aoltean16 (commented):
Increment and Decrement do seem like a good idea, but the other two can be incorporated in the fist 2, if you add increment / decrement step value as another parameter, with default of 1

TheBrambleShark:
This is true, but `number = number + 1` feels a bit awkward especially when we have support in the language for other operators.

Additionally, exposing these types as suggested will allow third party developers to contribute additional operators in the future. For example, I like working with nullable enabled, so having `nullableValue ??= foo` wrapped as an activity would be extremely valuable. Currently I have to do this as an if not null then assign, which doubles the amount of work I have to do for what should be a simple operation.

The IDE should make development easier, not harder.

PR #297, opened 2024-01-16
Make sure expressions end even if they left a quote open

aoltean16:
Such as:
VB: it does have even number of quotes
C#: it does not have unclosed multiline text or comment. Before the fix, when it had such situations, it would fail the whole validation for the file, making it impossible to pinpoint the actual error and fix it.
See https://uipath.atlassian.net/browse/STUD-67676 for more details

### DIFF
--- a/src/UiPath.Workflow/Validation/CSharpExpressionValidator.cs
+++ b/src/UiPath.Workflow/Validation/CSharpExpressionValidator.cs
@@ -16,9 +16,12 @@ namespace System.Activities.Validation;
 /// </summary>
 public class CSharpExpressionValidator : RoslynExpressionValidator
 {
-    private const string _valueValidationTemplate = "public static System.Linq.Expressions.Expression<System.Func<{0}>> CreateExpression{1}() => ({2}) => {3};//activityId:{4}";
-    private const string _delegateValueValidationTemplate = "{0}\npublic static System.Linq.Expressions.Expression<{1}<{2}>> CreateExpression{3}() => ({4}) => {5};//activityId:{6}";
-    private const string _referenceValidationTemplate = "public static {0} IsLocation{1}() => ({2}) => {3} = default({4});//activityId:{5}";
+    // This is used in case the expression does not properly close (e.g. missing quotes, or multiline comment not closed)
+    private const string _expressionEnder = "// */ // \"";
+
+    private const string _valueValidationTemplate = "public static System.Linq.Expressions.Expression<System.Func<{0}>> CreateExpression{1}()//activityId:{4}\n => ({2}) => {3}; {5}";
+    private const string _delegateValueValidationTemplate = "{0}\npublic static System.Linq.Expressions.Expression<{1}<{2}>> CreateExpression{3}()//activityId:{6}\n => ({4}) => {5}; {7}";
+    private const string _referenceValidationTemplate = "public static {0} IsLocation{1}()//activityId:{5}\n => ({2}) => {3} = default({4}); {6}";
 
     private static readonly Lazy<CSharpExpressionValidator> s_instance = new(() => new());
     public override string Language => CSharpHelper.Language;
@@ -54,10 +57,10 @@ public class CSharpExpressionValidator : RoslynExpressionValidator
     {
         var arrayType = types.Split(",");
         if (arrayType.Length <= 16) // .net defines Func<TResult>...Funct<T1,...T16,TResult)
-            return string.Format(_valueValidationTemplate, types, index, names, code, activityId);
+            return string.Format(_valueValidationTemplate, types, index, names, code, activityId, _expressionEnder);
 
         var (myDelegate, name) = CompilerHelper.DefineDelegate(types);
-        return string.Format(_delegateValueValidationTemplate, myDelegate, name, types, index, names, code, activityId);
+        return string.Format(_delegateValueValidationTemplate, myDelegate, name, types, index, names, code, activityId, _expressionEnder);
     }
 
     protected override string CreateReferenceCode(string types, string names, string code, string activityId, string returnType, int index)
@@ -65,7 +68,7 @@ public class CSharpExpressionValidator : RoslynExpressionValidator
         var actionDefinition = !string.IsNullOrWhiteSpace(types)
             ? $"Action<{string.Join(Comma, types)}>"
             : "Action";
-        return string.Format(_referenceValidationTemplate, actionDefinition, index, names, code, returnType, activityId);
+        return string.Format(_referenceValidationTemplate, actionDefinition, index, names, code, returnType, activityId, _expressionEnder);
     }
 
     protected override SyntaxTree GetSyntaxTreeForExpression(string expressionText) =>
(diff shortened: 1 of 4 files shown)


src/UiPath.Workflow.Runtime/Resources/SR.en.resx
```diff
@@ -3264,4 +3264,10 @@
   <data name="XamlElementExpectedAt" xml:space="preserve">
     <value>Element representation is expected at line ({0}, {1}).</value>
   </data>
+  <data name="InvalidFormattedExpressionCS" xml:space="preserve">
+    <value>Invalid expression. Please close the statement with " or the multiline comment with */</value>
+  </data>
+  <data name="InvalidFormattedExpressionVB" xml:space="preserve">
+    <value>Too many quotes. Please make sure the expression is valid!</value>
```
marius-bughiu: nit: Not technically correct. It might be too many, but it can also be a missing quote. Besides this, we should also make the message useful by providing some suggestions on how you can solve this error - something along the lines of: "The number of quotes in the expression is uneven. Please check that all double quotes have a matching end quote and that you have properly escaped any inner double quotes."

src/UiPath.Workflow/Validation/VBExpressionValidator.cs
```diff
@@ -42,6 +42,16 @@ protected VbExpressionValidator(HashSet<Assembly> referencedAssemblies = null)
 
     protected override string ActivityIdentifierRegex { get; } = "('activityId):(.*)";
 
+    internal override bool IsExpressionWellFormatted(ExpressionToValidate expression, List<ValidationError> errors)
+    {
+        if (!AreQuotesBalanced(expression.ExpressionText))
+        {
+            errors.Add(new ValidationError(SR.InvalidFormattedExpressionVB, expression.Activity));
```
marius-bughiu: `InvalidFormattedExpressionVB` is a very generic name, while the rule we are validating is very specific: even quotes. We might come across other scenarios in the future which we will validate and will have their own message, so I'd suggest we change this to something like `UnevenNumberOfQuotesVB`.

src/UiPath.Workflow/Validation/CSharpExpressionValidator.cs
```diff
@@ -75,4 +85,60 @@ protected override string CreateReferenceCode(string types, string names, string
...
 
     protected override string GetTypeName(Type type) => CompilerHelper.GetTypeName(type);
+
+    /// <summary>
+    /// Checks if there are multiline comments, and if they are properly closed
+    /// Checks if there are any multiline text statements, and if they are properly closed
+    /// </summary>
+    /// <param name="code"></param>
+    /// <returns></returns>
+    private static bool AreCommentsBalanced(string code)
```
marius-bughiu: nit: `code` -> `expression` / `expressionText` for consistency.

src/UiPath.Workflow/Validation/CSharpExpressionValidator.cs
```diff
@@ -75,4 +85,60 @@ protected override string CreateReferenceCode(string types, string names, string
...
+        else if (code[lastCharIndex] == '"')
+        {
+            isInsideQuotes = !isInsideQuotes;
+            if (multilineTextFaulted && !isInsideQuotes)
+            {
+                multilineTextFaulted = false;
+            }
+        }
+
+        return !(multilineCommentFaulted || multilineTextFaulted);
```
marius-bughiu: We should have different error messages for the two scenarios we are validating. Also in the unit tests we should validate them separately.

src/UiPath.Workflow/Validation/CSharpExpressionValidator.cs
```diff
@@ -75,4 +85,60 @@ protected override string CreateReferenceCode(string types, string names, string
...
+        bool multilineTextFaulted = false;
+        bool isInsideQuotes = false;
+
+        for (int i = 0; i < code.Length - 1; i++)
+        {
+            if (code[i] == '@' && !isInsideQuotes)
+            {
+                multilineTextFaulted = true;
+            }
+            else if (code[i] == '"')
```
marius-bughiu: This doesn't account for escape sequences. If you have a `\"` in your string, `isInsideQuotes` will become `false` and using `@` or `/*` after that point will result in false positives.

src/UiPath.Workflow/Validation/CSharpExpressionValidator.cs
```diff
@@ -75,4 +85,60 @@ protected override string CreateReferenceCode(string types, string names, string
...
+            {
```
marius-bughiu: Besides the escape issue above, there are some more advanced scenarios which are not really handled. Two that come to mind are:

- raw string literals
```
"""
    Lorem ipsum "dolor" sit amet,
        consectetur adipiscing elit.
"""
```

- string interpolation, like `$"Hi, { "foo" + "bar" }"` - where who knows what the users will do...

marius-bughiu (changes requested):

src/UiPath.Workflow/Validation/CSharpExpressionValidator.cs
aoltean16: The prevalidate does not make sure this is a true valid expression, but rather that it will not break my compilation of all expressions, resulting in the bug described above. Both the scenarios provided here will cause an isolated error, which is enough to pinpoint the actual error. The problem is when your expression actually breaks the compilation and it cannot determine properly where the issue is.

src/UiPath.Workflow/Validation/CSharpExpressionValidator.cs
aoltean16: i'd say it's debatable whether we want to be that granular or not. as they both cause about the same issue: the given expression is invalid cause you forgot to close it properly (comment or quote)

src/UiPath.Workflow/Validation/CSharpExpressionValidator.cs
aoltean16: damn, nice catch!

src/UiPath.Workflow/Validation/CSharpExpressionValidator.cs
marius-bughiu: Won't this generate false positives? Say I have this expression:

```
"""
    This is a quote: "
"""
```

src/UiPath.Workflow/Validation/CSharpExpressionValidator.cs
```diff
@@ -16,9 +16,12 @@ namespace System.Activities.Validation;
 /// </summary>
 public class CSharpExpressionValidator : RoslynExpressionValidator
 {
-    private const string _valueValidationTemplate = "public static System.Linq.Expressions.Expression<System.Func<{0}>> CreateExpression{1}() => ({2}) => {3};//activityId:{4}";
-    private const string _delegateValueValidationTemplate = "{0}\npublic static System.Linq.Expressions.Expression<{1}<{2}>> CreateExpression{3}() => ({4}) => {5};//activityId:{6}";
-    private const string _referenceValidationTemplate = "public static {0} IsLocation{1}() => ({2}) => {3} = default({4});//activityId:{5}";
+    // This is used in case the expression does not properly close (e.g. missing quotes, or multiline comment not closed)
+    public const string ExpressionEnder = "// */ // \"";
```
marius-bughiu: Should be made `private`, and ideally would follow the same naming convention as the constants below (`_camelCase`). The naming part goes for it's counterpart in `VBExpressionValidator` as well.

marius-bughiu (approved):
Nice solution. Only had one comment to add for naming consistency, but otherwise it's good to go from my perspective.

PR #298, opened 2024-01-19
Activity RefreshActitivityTree

AndreiTimisescu:
A client may change properties at runtime (while execution is paused), operation which requires a new call to CacheMetadata.
The Activity.RefreshActivityTree extension method provides this functionality.

https://uipath.atlassian.net/wiki/spaces/STUD/pages/87546724494/Hot+Reload+Investigation

### DIFF
--- a/src/UiPath.Workflow/Debugger/ActivityExtensions.cs
+++ b/src/UiPath.Workflow/Debugger/ActivityExtensions.cs
@@ -1,0 +1,74 @@
+﻿using System;
+using System.Activities;
+using System.Activities.Runtime;
+using System.Activities.Tracking;
+using System.Activities.Validation;
+using System.Collections.Generic;
+using System.Linq;
+
+namespace UiPath.Workflow.Debugger;
+
+public static class ActivityExtensions
+{
+    /// <summary>
+    /// A client may change properties at runtime (while execution is paused),
+    /// and so, based on the activity internal logic, a new call to CacheMetadata may be required.
+    /// This extension method will update the activity state.
+    /// ATTENTION - calling this method will generate a new activity.Id
+    /// </summary>
+    /// <returns>a list of validation errors</returns>
+    public static IList<ValidationError> RefreshActivityTree(this Activity activity)
+    {
+        if (activity == null)
+            return Array.Empty<ValidationError>();
+
+        ClearCacheIds(activity);
+        activity.ClearCachedInformation();
+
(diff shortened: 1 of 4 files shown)


src/UiPath.Workflow.Runtime/Tracking/TrackingProvider.cs
```diff
@@ -324,6 +324,12 @@ private bool PostTrackingRecord(TrackingParticipant participant, RuntimeTracking
                 if (preparedRecord != null)
                 {
                     IAsyncResult result = participant.BeginTrack(preparedRecord, _timeoutHelper.RemainingTime(), PrepareAsyncCompletion(trackingCompleteCallback), this);
+
+                    if (preparedRecord is ActivityScheduledRecord record)
```
gabriela-lungu-uip: this affects existing workflows and functionalities

gabriela-lungu-uip (changes requested):

src/UiPath.Workflow.Runtime/Tracking/TrackingProvider.cs
AndreiTimisescu: It should do nothing as nothing changes on the regular path. But you are right, we may call this method only when needed.

gabriela-lungu-uip (approved):
UiPath.Workflow.Runtime code looks ok. This logic implemented in this PR is only called during Debug, so production/deployed workflows should not be impacted by the changes.
I did not review Debugger logic implemented in UiPath.Workflow

src/UiPath.Workflow/Debugger/ActivityExtensions.cs
```diff
@@ -0,0 +1,74 @@
...
+public static class ActivityExtensions
+{
+    /// <summary>
+    /// A client may change properties at runtime (while execution is paused),
+    /// and so, based on the activity internal logic, a new call to CacheMetadata may be required.
+    /// This extension method will update the activity state.
+    /// ATTENTION - calling this method will generate a new activity.Id
+    /// </summary>
+    /// <returns>a list of validation errors</returns>
+    public static IList<ValidationError> RefreshActivityTree(this Activity activity)
```
aoltean16: Please rename this to something more meaningful, like RefreshActivityArguments. This way we "isolate" more what we're doing here, in case we change it in the future.

aoltean16 (approved):

PR #299, opened 2024-01-29
Use AssemblyReferences instead of AssemblyNames for validation

andrei-ungureanu-uipath:
There are two types of validation
- For the whole workflow
- For a single expression

The first one creates the compilation by using a list of `AssemblyReferences`. The second one, uses a list of `AssemblyNames` and **it either resolves them or retrieves the values from a cache**. If the platform that requires the validation already has an `AssemblyReference`, **it should be able to forward without relying on CoreWF to resolve it**. 

This PR replaces the list of `AssemblyNames` with `AssemblyReferences` and uses the same approach as the full workflow validation to resolve the `Assemblies`.

### DIFF
--- a/src/UiPath.Workflow/Activities/DesignerHelperImpl.cs
+++ b/src/UiPath.Workflow/Activities/DesignerHelperImpl.cs
@@ -3,6 +3,7 @@
 
 using Microsoft.CodeAnalysis;
 using Microsoft.CodeAnalysis.Diagnostics;
+using Microsoft.PowerFx.Core.Public.Values;
 using Microsoft.VisualBasic.Activities;
 using System.Activities.ExpressionParser;
 using System.Activities.Expressions;
@@ -34,9 +35,7 @@ internal abstract class DesignerHelperImpl
 
     protected abstract ExpressionCompiler Compiler { get; }
 
-    public abstract JitCompilerHelper CreateJitCompilerHelper(string expressionText, HashSet<AssemblyName> references,
-        HashSet<string> namespaces);
-
+    public abstract JitCompilerHelper CreateJitCompilerHelper(string expressionText, HashSet<AssemblyReference> references, HashSet<string> namespaces);
 
     public Activity CreatePrecompiledValue(Type targetType,
         string expressionText,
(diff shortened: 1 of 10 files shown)


src/UiPath.Workflow/Microsoft/VisualBasic/Activities/VisualBasicDesignerHelper.cs
```diff
@@ -143,7 +150,16 @@ public static Task<CompiledExpressionResult> CreatePrecompiledValueAsync(Type ta
             environment, out returnType, out compileError, out vbSettings);
     }
 
-    public static Task<CompiledExpressionResult> CreatePrecompiledReferenceAsync(Type targetType, string expressionText, IEnumerable<string> namespaces, IEnumerable<string> assemblies, LocationReferenceEnvironment environment)
+    public static Activity CreatePrecompiledVisualBasicReference(Type targetType, string expressionText,
```
aoltean16, line 153: Please go with `CreatePrecompiledReferenceAsync`. The language name in the method name does us no good

src/UiPath.Workflow/Activities/DesignerHelperImpl.cs
```diff
@@ -201,35 +188,23 @@ public Task<CompiledExpressionResult> RecompileValueAsync(ActivityWithResult rVa
...
-        {
-            foreach (var assm in referencedAssemblies)
-            {
-                if (assm != null)
-                {
-                    assembliesSet.Add(new AssemblyName(assm));
-                }
-            }
-        }
-
```
aoltean16, line 232: Where did this go?

aoltean16 (commented):
Should ask SWeb to take a look as well.

src/UiPath.Workflow/Microsoft/VisualBasic/Activities/VisualBasicDesignerHelper.cs
andrei-ungureanu-uipath: This is not a new method. Above it there is also a `CreatePrecompiledVisualBasicReference` (i'm also not a fan of the name). I added no new methods in the `DesignerHelper` files, only method overloads from `IEnumerable<string>` to `IEnumerable<AssemblyReference>`

src/UiPath.Workflow/Activities/DesignerHelperImpl.cs
andrei-ungureanu-uipath: I moved them inline a [few lines above](https://github.com/UiPath/CoreWF/pull/299/files#diff-b9aa237e8abd9398a613f1fbebd598a3100bb4f101cc0327a7d06b23fc8db50cR205-R208).

aoltean16: haha, my bad, i did notice those.

src/UiPath.Workflow/Microsoft/VisualBasic/Activities/VisualBasicDesignerHelper.cs
aoltean16: that was my point. instead of overloads, we could've used the proper name :)

aoltean16 (approved):

PR #305, opened 2024-03-04
test: compilation should work

mirceaarhip:


src/UiPath.Workflow/Activities/ScriptingJitCompiler.cs
```diff
@@ -86,8 +86,25 @@ public override LambdaExpression CompileExpression(ExpressionToCompile expressio
...
+
+        TextExpressionCompilerResults TryBuild()
+        {
+            try
+            {
+                return ScriptingAotCompiler.BuildAssembly(finalCompilation, compilation.ScriptClass.Name, collectibleAlc);
+            }
+            catch
+            {
+                var errorResults = ScriptingAotCompiler.GetCompilerResults(compilation);
```
aoltean16: Please do not use the `compilation` variable for getting diagnostics, as its purpose differs from the actual script (`finalCompilation`). You could throw a rather generic exception or so.

aoltean16 (approved):

PR #307, opened 2024-03-05
Add support for distinguishing which Parallel/Pick parallelism branch an activity is executing in

gabriela-lungu-uip:
This feature introduces a context.GetCurrentParallelId() queryable from an activity,
which should identify a "parallelism" branch on which the activity executes.

Only containers which schedule concurrently (Parallel, ParallelForEach, Pick)
would induce a change of the background parallel id.

The initial parallel id, before using the above containers, is null.


src/UiPath.Workflow.Runtime/ActivityContext.cs
```diff
@@ -86,7 +86,11 @@ public WorkflowDataContext DataContext
         }
     }
```
vuplea: this feature can be done in an extension class; it's better not to mess with core classes Activity/ActivityInstance/ActivityContext if not necessary.

vuplea: My suggestion in a separate file, including some naming changes is:
```
+namespace UiPath.Workflow.Runtime.ParallelTracking
+{
+    public static class ParallelTrackingExtensions
+    {
+        public const string BranchIdPropertyName = "__ParallelBranchId";
+
+        public static ActivityInstance MarkNewParallelBranch(this ActivityInstance instance)
+        {
+            var properties = new ExecutionProperties(null, instance, instance.PropertyManager);
+            var parentId = properties.Find(BranchIdPropertyName) as string;
+            properties.Add(BranchIdPropertyName, $"{parentId}.{Guid.NewGuid():N}", true, false);
+            return instance;
+        }
+
+        public static string GetCurrentParallelBranchId(this ActivityInstance instance) =>
+            instance.PropertyManager?.GetPropertyAtCurrentScope(BranchIdPropertyName) as string;
+    }
```
we can include the extensions on ActivityContext also.

src/UiPath.Workflow.Runtime/ActivityInstance.cs
```diff
@@ -402,7 +402,16 @@ internal string OwnerName
...
 
     [DataMember(EmitDefaultValue = false)]
-    public Version ImplementationVersion { get; internal set; }
+    public Version ImplementationVersion { get; internal set; }
+
+    public void AddAutomationTrackerId()
+    {
+        var properties = new ExecutionProperties(null, this, PropertyManager);
+        var existing = properties.Find("BPOId") as string;
+        properties.Add("BPOId", (existing + "-" + Guid.NewGuid().ToString("N")).Trim('-'), true, false);
```
vuplea: `.` is the separator for activityIds, so I was thinking to use the same?

gabriela-lungu-uip: done

src/UiPath.Workflow.Runtime/ActivityContext.cs
gabriela-lungu-uip: done

PR #317, opened 2024-04-18
STUD-69419 add flowchart interfaces

andrei-balint:
Define flowchart and flow nodes interfaces.
The purpose is to allow a designer host app to handle multiple flowchart implementations.

### DIFF
--- a/src/UiPath.Workflow.Runtime/Statements/FlowNode.cs
+++ b/src/UiPath.Workflow.Runtime/Statements/FlowNode.cs
@@ -1,9 +1,11 @@
 // This file is part of Core WF which is licensed under the MIT license.
 // See LICENSE file in the project root for full license information.
 
+using System.Activities.Statements.Interfaces;
+
 namespace System.Activities.Statements;
 
-public abstract class FlowNode
+public abstract class FlowNode: IFlowNode
 {
     private Flowchart _owner;
     private int _cacheId;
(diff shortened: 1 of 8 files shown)


src/UiPath.Workflow.Runtime/Statements/FlowNode.cs
```diff
@@ -60,4 +62,11 @@ internal void GetChildActivities(ICollection<Activity> children)
     }
 
     internal abstract void GetConnectedNodes(IList<FlowNode> connections);
+
+    IEnumerable<IFlowNode> IFlowNode.GetConnectedNodes()
+    {
+        var result = new List<FlowNode>();
+        GetConnectedNodes(result);
+        return result;
+    }
```
mihainradu, line 71: why?

andrei-balint: There is reflection code in Studio and I would like to take the opportunity to add a proper method.

mihainradu: as aa warning, this is internal and I would expect to be able to rename it

andrei-balint: unfortunately Studio is using it with reflection as there are no contracts available

andrei-balint:
will be declared by design time assembly only

PR #319, opened 2024-04-29
Fix resolving coded types while typing in ETB STUD-69826

gabrielavaduva:
This resolves the issue where typing in an expressionTextbox expecting a return type of a type with a generic argument type comes from an assembly that is in another ALC, than his parent.
E.g. List<MyCodedType> as a returnType in an expressionTextbox, it would fail, as reflection cannot resolve at-once a type from string, if atleast one of its children come from another ALC, so we resolve type-by-type and compose the resulting type in the end.

## Jira
https://uipath.atlassian.net/browse/STUD-69826


AlvinStanescu (approved):

src/UiPath.Workflow/Activities/ExpressionCompiler.cs
```diff
@@ -35,6 +35,19 @@ protected static Assembly GetAssemblyForType(ITypeSymbol type)
         protected static Type GetSystemType(ITypeSymbol typeSymbol, Assembly assembly)
         {
             StringBuilder typeName = new();
+
+            if (typeSymbol is INamedTypeSymbol namedTypeSymbol)
+            {
+                if (namedTypeSymbol.IsGenericType)
+                {
+                    var arguments = namedTypeSymbol.TypeArguments.Select(t => GetSystemType(t, GetAssemblyForType(t))).ToArray();
```
AlvinStanescu: Also test if this is a generic of a type in coded (e.g.) `IEnumerable<CodedType>`

gabrielavaduva: Tested it now, it works.

src/Test/TestCases.Workflows/ExpressionTests.cs
```diff
@@ -566,4 +566,34 @@ public void VB_IsCaseInsensitive()
...
+        sequence.Variables.Add(new Variable<List<ArgumentDirection>>("var1"));
+        sequence.Activities.Add(new Assign
+        {
+            To = new OutArgument<List<ArgumentDirection>>(new CSharpReference<List<ArgumentDirection>>("var1")),
+            Value = new InArgument<List<ArgumentDirection>>(new CSharpValue<List<ArgumentDirection>>("null"))
+        });
+
+        var result = ActivityValidationServices.Validate(sequence, _useValidator);
+        result.Errors.ShouldBeEmpty();
+    }
```
aoltean16: These do not touch the actual code changed below. Please check `VB_CreatePrecompiedValueAsync_ProperlyIdentifiesVariables` as an example, as this is the code path for which the fix below tries to apply.

src/UiPath.Workflow/Activities/ExpressionCompiler.cs
```diff
@@ -35,6 +35,19 @@ protected static Assembly GetAssemblyForType(ITypeSymbol type)
...
+                    var fullyQualifiedName = $"{namedTypeSymbol.ContainingNamespace.ToDisplayString()}.{namedTypeSymbol.MetadataName}";
+                    var genericDefinitionType = assembly.GetType(fullyQualifiedName);
+
+                    return genericDefinitionType.MakeGenericType(arguments);
+                }
+            }
```
aoltean16: This code is treated 10 lines below, in method `GetTypeName(ITypeSymbol typeSymbol)`. It does the same check as this, and properly identifies both arrays and generic arguments for any type.

aoltean16 (changes requested):
Please link the jira issue, and let's see what the actual problem is, as it is not clear from the unit tests

aoltean16 (approved):

PR #321, opened 2024-05-07
Use argument display name for validation message when available instead of its name STUD-68550

pawelpopielarski:
https://uipath.atlassian.net/browse/STUD-68550

Connections in libraries feature is generating properties with names, which we do not want to display during validation.
We want to use display names, which we set during compilation, hence this change.

### DIFF
--- a/src/UiPath.Workflow.Runtime/RuntimeArgument.cs
+++ b/src/UiPath.Workflow.Runtime/RuntimeArgument.cs
@@ -78,6 +78,8 @@ public sealed class RuntimeArgument : LocationReference
 
     public ArgumentDirection Direction { get; private set; }
 
+    public string DisplayName => _bindingProperty?.DisplayName ?? Name;
+
     public bool IsRequired { get; private set; }
 
     public ReadOnlyCollection<string> OverloadGroupNames
--- a/src/UiPath.Workflow.Runtime/RuntimeArgument.cs
+++ b/src/UiPath.Workflow.Runtime/RuntimeArgument.cs
@@ -78,7 +78,7 @@ public sealed class RuntimeArgument : LocationReference
 
     public ArgumentDirection Direction { get; private set; }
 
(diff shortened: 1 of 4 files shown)


src/UiPath.Workflow.Runtime/UiPath.Workflow.Runtime.csproj
```diff
@@ -26,7 +26,7 @@
 	<ProjectCapability Include="ConfigurableFileNesting" />
 	<ProjectCapability Include="ConfigurableFileNestingFeatureEnabled" />
   </ItemGroup>
-  <Target Name="CopyProjectReferencesToPackage" DependsOnTargets="BuildOnlySettings;ResolveReferences" Condition="$(Configuration)=='Release'">
```
aoltean16: Not related to this PR. Let's do this separately :)

aoltean16 (commented):
Please make unit tests showcasing old and new behavior (no display name, and a display name)

aoltean16 (approved):
Looks good from my side, but since it touches the runtime, we need to have the approval from @gabriela-lungu-uip too, before proceeding.

src/UiPath.Workflow.Runtime/RuntimeArgument.cs
```diff
@@ -78,6 +78,8 @@ internal static IComparer<RuntimeArgument> EvaluationOrderComparer
 
     public ArgumentDirection Direction { get; private set; }
 
+    public string DisplayName => _bindingProperty?.DisplayName ?? Name;
```
gabriela-lungu-uip: this can be a breaking change. Being a public property on a public type, NewtonsoftJson serialization will serialize it and if the workflow is resumed on an older Robot it will fail. Mark it internal instead

gabriela-lungu-uip (changes requested):

gabriela-lungu-uip (approved):

PR #322, opened 2024-05-07
Pack System.Xaml also in debug

pawelpopielarski:
I dont know why it was packed only in Release, 
I lost an hour trying to figure out what is wrong when building Studio with a local version of CoreWF.

When using a locally built CoreWF package, Studio build gives a whole lot of strange errors because `System.Xaml.dll` is missing from net6 lib in package


aoltean16 (approved):

bogdan-toma-uipath:
let's wait robot input here.

gabriela-lungu-uip (approved):

PR #323, opened 2024-05-09
Handle return type of 'null' expressions

andrei-ungureanu-uipath:
When a precompiled value is created, determining the `ReturnType` of an expression may throw an exception for certain values like `null` and `() => null` because there needs to be an implicit cast on the `SemanticModel` to figure out the type. In these scenarios, the `Type` property will be null and we should fallback to the `ConvertedType` [property](https://learn.microsoft.com/en-us/dotnet/api/microsoft.codeanalysis.typeinfo.convertedtype?view=roslyn-dotnet-4.7.0#microsoft-codeanalysis-typeinfo-convertedtype) as this one is populated when it cannot be inferred without casting.

### DIFF
--- a/src/Test/TestCases.Workflows/XamlTests.cs
+++ b/src/Test/TestCases.Workflows/XamlTests.cs
@@ -11,6 +11,7 @@ using System.Activities.Statements;
 using System.Activities.Validation;
 using System.Activities.XamlIntegration;
 using System.Collections.Generic;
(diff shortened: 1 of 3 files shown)


aoltean16 (approved):

aoltean16:
These are both design-only features, as they're only called when typing into an expression textbox

PR #325, opened 2024-06-06
add IRedirectDebugTarget

mihainradu:
This will be used to facilitate identifing objects that are more meaningfull in a designer as a breakpoint target

eg: in a Diagram/flow node an implementation/runtime activity that implements this interface will trigger the tracker/debuger
then the debugger will do the check:
```
if (targetObj is IRedirectDebugTarget redirection)
   targetObj = redirection.Target
   
   ...
     return GetIdRef(targetObj)
   ```


andrei-balint (approved):
🚀

PR #328, opened 2024-06-17
process locations jit

AndreiSingeorzan:
Make CompiledExpressionInvoker.ProcessLocationReferences lazy, only when validating.

There's a project (from a client) with 1400 global constants. Because all the global constants (and vars) are added into a global scope for data manager, and because you can theoretically (from a WF point of view) use any other variable in a variable, there's lots of LocationReference instances being created, each with associated state (e.g. autogenerated args). Each of the 1400 constants has location references to the other 1399 constants. 
This causes a lot of memory pressure (non leak). Project from the issue (the actual attachment is in the issue clone) uses around 3 GB of after opening and waiting for a while.

With the fix, on my machine, it climbs to about 1.2 GB before going back down to about 600MB.

This will be followed by a PR in studio which uses the ForceExpressionCache when validating the data manager global scope sequence.i

Studio PR: https://github.com/UiPath/Studio/pull/19913

https://uipath.atlassian.net/browse/STUD-70469

### DIFF
--- a/src/UiPath.Workflow.Runtime/Expressions/CompiledExpressionInvoker.cs
+++ b/src/UiPath.Workflow.Runtime/Expressions/CompiledExpressionInvoker.cs
@@ -22,6 +22,7 @@ public class CompiledExpressionInvoker
     private readonly IList<LocationReference> _locationReferences;
     private CodeActivityMetadata _metadata;
     private CodeActivityPublicEnvironmentAccessor _accessor;
+    private bool _locationsInitialized;
 
     public CompiledExpressionInvoker(ITextExpression expression, bool isReference, CodeActivityMetadata metadata)
     {
@@ -40,7 +41,10 @@ public class CompiledExpressionInvoker
 
         _metadataRoot = metadata.Environment.Root;
 
-        ProcessLocationReferences();
+        if (!metadata.Environment.IsValidating)
+        {
+            ProcessLocationReferences();
+        }
     }
 
     public object InvokeExpression(ActivityContext activityContext)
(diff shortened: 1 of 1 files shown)


src/UiPath.Workflow.Runtime/Expressions/CompiledExpressionInvoker.cs
```diff
@@ -244,6 +258,8 @@ private void ProcessLocationReferences()
                 CreateRequiredArguments(requiredLocationNames);
             }
         }
+
+        _locationsInitialized = true;
```
gabriela-lungu-uip: _locationsInitialized = true; should be set at the beginning of the method to ensure the previous behavior

src/UiPath.Workflow.Runtime/Expressions/CompiledExpressionInvoker.cs
```diff
@@ -40,7 +41,10 @@ public CompiledExpressionInvoker(ITextExpression expression, bool isReference, C
 
         _metadataRoot = metadata.Environment.Root;
 
-        ProcessLocationReferences();
+        if (!metadata.Environment.IsValidating)
```
gabriela-lungu-uip, line 44: IsValidating is default false;
I can be set to true only in InternalActivityValidationServices constructor which is only called from Studio code either directly or indirectly. It is not called from anywhere in the production Robot code or Activities repository production code.
So the changes described in this PR do not affect the Robot.
In regards to the problem being fixed, this should be assessed by @aoltean16 or someone else from the Studio team

AndreiSingeorzan: I've already asked for review from Alex, but as a general principle, this can also be reproduced in a simple unit test in this project (create a model item with 1400 variables and validate it with/without the ForceExpressionCache flag and observe the memory).

aoltean16 (approved):

andrei-balint (approved):

PR #329, opened 2024-07-04
use dedicated flag for skipping calculation of location references

AndreiSingeorzan:
Use a dedicated flag to skip calculating location references, instead of the existing IsValidating flag. Reason is the scenario from the below issue.

This new flag is only set to true for data manager global scope (where one global variable cannot reference another one (and same goes for global constants)).

I've retested the large client project, and the memory pressure fix for it remains in place (once studio will set the new flag to true for data manager).

https://uipath.atlassian.net/browse/STUD-70876


gabriela-lungu-uip:
the changes in this PR do not affect the Robot runtime and apply only for validation

PR #333, opened 2024-07-29
feat: add dateonly converter

danield20:
Adding datateonly and timeonly convertors (taken from .net 8) to coreWF

https://uipath.atlassian.net/browse/SW-14266

### DIFF
--- a/src/System.Xaml/System/Windows/Markup/TypeConverterHelper.cs
+++ b/src/System.Xaml/System/Windows/Markup/TypeConverterHelper.cs
@@ -244,6 +244,14 @@ namespace System.Windows.Markup
             {
                 converterType = typeof(DateTimeConverter2);
             }
+            else if (typeof(DateOnly).IsAssignableFrom(type))
+            {
+                typeConverter = typeof(DateOnlyConverter2);
+            }
+            else if (typeof(TimeOnly).IsAssignableFrom(type))
+            {
+                typeConverter = typeof(TimeOnlyConverter2);
+            }
 
             return converterType;
         }
@@ -338,6 +346,14 @@ namespace System.Windows.Markup
             else if (ReflectionHelper.IsNullableType(type))
             {
                 typeConverter = new System.ComponentModel.NullableConverter(type);
+            }
+            else if (type == typeof(DateOnly))
+            {
+                typeConverter = new DateOnlyConverter2();
+            }
+            else if (type == typeof(TimeOnly))
+            {
+                typeConverter = new TimeOnlyConverter2();
             }
 
             return typeConverter;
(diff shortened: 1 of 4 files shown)


src/System.Xaml/System/Windows/Markup/TypeConverterHelper.cs
```diff
@@ -244,6 +244,10 @@ internal static Type GetCoreConverterTypeFromCustomType(Type type)
             {
                 converterType = typeof(DateTimeConverter2);
             }
+            else if (typeof(DateOnly).IsAssignableFrom(type))
```
mirceaarhip, line 247: what about TimeOnly

src/System.Xaml/System/Xaml/Replacements/DateOnlyConvertor.cs
```diff
@@ -0,0 +1,97 @@
...
+using System.ComponentModel.Design.Serialization;
+using System.Diagnostics.CodeAnalysis;
+using System.Globalization;
+
+namespace System.ComponentModel
+{
+    /// <summary>
+    /// Provides a type converter to convert <see cref='System.DateOnly'/> objects to and from various other representations.
+    /// </summary>
+    public class DateOnlyConverter : TypeConverter
```
mirceaarhip: let's use a different name so we won't  have conflicts with the real type

mirceaarhip: an the namespace

mirceaarhip (changes requested):

src/System.Xaml/System/Xaml/Replacements/TimeOnlyConverter2.cs
```diff
@@ -0,0 +1,100 @@
...
+using System.ComponentModel.Design.Serialization;
+using System.Diagnostics.CodeAnalysis;
+using System.Globalization;
+
+namespace System.Xaml.Replacements
+{
+    /// <summary>
+    /// Provides a type converter to convert <see cref='System.TimeOnly'/> objects to and from various other representations.
+    /// </summary>
+    public class TimeOnlyConverter2 : TypeConverter
```
mirceaarhip: can this be internal? Also not convinced this is a good name.

mirceaarhip (changes requested):

src/System.Xaml/System/Xaml/Replacements/TimeOnlyConverter2.cs
danield20: changed them to internal. The name is similar to the ones previously used eg. `DateTimeConverter2`

mirceaarhip (approved):

aoltean16 (approved):

PR #343, opened 2024-11-05
Add new extension for cases where a required extension is missing ROBO-4220

thecristidima:
# Description 
In case of a missing required extensions we throw a ValidationException. Users might want to handle those cases explicitly and they would need to know the type of the missing extension.

To achieve this, we added ExtensionRequiredException which inherits ValidationException and has a single property RequiredExtensionType.

# Jira
https://uipath.atlassian.net/browse/ROBO-4220

### DIFF
--- a/src/Test/TestCases.Runtime/MissingRequiredExtensionTests.cs
+++ b/src/Test/TestCases.Runtime/MissingRequiredExtensionTests.cs
@@ -29,8 +29,9 @@ namespace TestCases.Runtime
             }
             else
             {
-                exception.ShouldBeOfType<ExtensionRequiredException>();
-                ((ExtensionRequiredException)exception).RequiredExtensionType.ShouldBe(typeof(StringBuilder));
+                exception.ShouldBeOfType<ValidationException>();
+                exception.InnerException.ShouldBeOfType<ExtensionRequiredException>();
+                ((ExtensionRequiredException)exception.InnerException).RequiredExtensionType.ShouldBe(typeof(StringBuilder));
             }
         }
 
--- a/src/Test/TestCases.Runtime/MissingRequiredExtensionTests.cs
+++ b/src/Test/TestCases.Runtime/MissingRequiredExtensionTests.cs
@@ -5,7 +5,6 @@ using Shouldly;
 using System.Activities;
 using System.Activities.Statements;
 using System.Text;
-using Test.Common.TestObjects.Activities;
 using Xunit;
 
 namespace TestCases.Runtime
@@ -23,6 +22,18 @@ namespace TestCases.Runtime
 
             var ex = Assert.Throws<ExtensionRequiredException>(instance.Run);
             ex.RequiredExtensionType.ShouldBe(typeof(StringBuilder));
+        }
+
+        private class MissingRequiredExtension<T> : NativeActivity
+        {
+            protected override void CacheMetadata(NativeActivityMetadata metadata)
+            {
+                metadata.RequireExtension(typeof(T));
+            }
+
+            protected override void Execute(NativeActivityContext context)
+            {
+            }
         }
     }
 }
--- a/src/Test/TestCases.Runtime/MissingRequiredExtensionTests.cs
+++ b/src/Test/TestCases.Runtime/MissingRequiredExtensionTests.cs
@@ -1,0 +1,28 @@
+﻿// This file is part of Core WF which is licensed under the MIT license.
+// See LICENSE file in the project root for full license information.
+
+using Shouldly;
+using System.Activities;
+using System.Activities.Statements;
+using System.Text;
+using Test.Common.TestObjects.Activities;
+using Xunit;
+
+namespace TestCases.Runtime
+{
+    public class MissingRequiredExtensionTests
+    {
+        [Fact]
+        public void RunningActivityWithUnregisteredRequiredExtensionShouldThrow()
+        {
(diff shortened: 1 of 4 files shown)


src/Test/TestObjects/Activities/TestReadLine.cs
```diff
@@ -229,4 +229,16 @@ protected override void GetActivitySpecificTrace(TraceGroup traceGroup)
             traceGroup.Steps.Add(new UserTrace(WaitReadLine<T>.AfterWait));
         }
     }
+
```
mihainradu: why didn't you put this in the test class?

src/Test/TestCases.Runtime/MissingRequiredExtensionTests.cs
```diff
@@ -0,0 +1,39 @@
...
+
+using Shouldly;
+using System.Activities;
+using System.Activities.Statements;
+using System.Text;
+using Xunit;
+
+namespace TestCases.Runtime
+{
+    public class MissingRequiredExtensionTests
```
alexmustata19, line 12: Maybe we should add also a test to verify that no exception is thrown when a type is successfully resolved. I haven't seen one such test from what I searched.

src/Test/TestCases.Runtime/MissingRequiredExtensionTests.cs
```diff
@@ -0,0 +1,60 @@
...
+                if (_registerExtension)
+                {
+                    metadata.AddDefaultExtensionProvider(() => new StringBuilder());
+                }
+            }
+
+            protected override void Execute(NativeActivityContext context)
+            {
+            }
+        }
```
alexmustata19, line 60:
```suggestion
private class ActivityWithRequiredExtension<T> : NativeActivity where T : class, new()
{
    private readonly bool _registerExtension;

    public ActivityWithRequiredExtension(bool registerExtension)
    {
        _registerExtension = registerExtension;
    }

    protected override void CacheMetadata(NativeActivityMetadata metadata)
    {
        metadata.RequireExtension(typeof(T));
        if (_registerExtension)
        {
            metadata.AddDefaultExtensionProvider(() => new T());
        }
    }

    protected override void Execute(NativeActivityContext context)
    {
    }
}
```
`ActivityWithRequiredExtension` can be kept generic if `T` has constrains to be a reference type and to have a parameterless constructor.

alexmustata19: Line `21` will become:
`Activities = { new ActivityWithRequiredExtension<StringBuilder>(registerExtension) }`

thecristidima: But we only use it in this single test and we're not interested in testing different extension types, so I'd keep it simple and hardcode a type.

alexmustata19 (approved):

src/UiPath.Workflow.Runtime/Hosting/WorkflowInstanceExtensionCollection.cs
```diff
@@ -118,7 +118,7 @@ internal WorkflowInstanceExtensionCollection(Activity workflowDefinition, Workfl
                 {
                     if (!TypeHelper.ContainsCompatibleType(allExtensionTypes, requiredType))
                     {
-                        throw FxTrace.Exception.AsError(new ValidationException(SR.RequiredExtensionTypeNotFound(requiredType.ToString())));
+                        throw FxTrace.Exception.AsError(new ExtensionRequiredException(requiredType));
```
vuplea: Introducing a breaking change for the most minimal need is not paying off at all.  I also think that not using ValidationException breaks things in general (search for references for this type in the framework, and you might find custom logic for it).
We need a way to avoid breaking changes completely for this.

vuplea (changes requested):

src/UiPath.Workflow.Runtime/Hosting/WorkflowInstanceExtensionCollection.cs
mihainradu: One way around this would be to make the new exception an InnerException (also it may need to be serializable)

src/UiPath.Workflow.Runtime/ExtensionRequiredException.cs
mihainradu, line 1: see SourceExpressionException for implemeting serializable with ISerializable

src/UiPath.Workflow.Runtime/ExtensionRequiredException.cs
```diff
@@ -0,0 +1,42 @@
...
+// See LICENSE file in the project root for full license information.
+
+namespace System.Activities;
+
+[Serializable]
+public class ExtensionRequiredException : Exception
+{
+    private const string RequiredExtensionTypeName = "requiredExtensionType";
+
+    public string RequiredExtensionTypeFullName { get; }
```
mihainradu: if you fwd this to the Data member you won't need to handle serialization as it is already handled

mihainradu: Data[RequiredExtensionTypeName]

thecristidima: Done

mihainradu (approved):

PR #346, opened 2024-12-17
[ROBO-4338] Add Owner field to BookmarkInfo

daniel-stanciu:


src/Directory.Build.props
```diff
@@ -1,6 +1,6 @@
 <Project>
     <PropertyGroup>
-        <Version>6.0.0</Version>
```
mihainradu: revert

mihainradu:
pls add tests with persistIdle arg.GetBookmarks, and directly accessing workflowapplication.GetBookmarks

mihainradu:
maybe its worth to modify ann make sure we receive BookmarkInfo for all bookmarks ... even emptynames... if that seems problematc we could add a new method... the purpose here would be to also detect bookmarks like the ones from Delay activity

src/UiPath.Workflow.Runtime/Hosting/BookmarkInfo.cs
```diff
@@ -39,6 +41,12 @@ public BookmarkScopeInfo ScopeInfo
         private set => _scopeInfo = value;
     }
 
+    public Activity Owner
```
mihainradu, line 44: Add tests this is set ok... also after a suspension

src/Test/WorkflowApplicationTestExtensions/WorkflowApplicationTestSamples.cs
mihainradu, line 1: revert

mihainradu (approved):

PR #347, opened 2024-12-18
robo-4382 Introduce option to opt-out of bookmark creation in Delay statement

mciureanu:
Current behaviour: the job never gets resumed from Orchestrator when using Delay.

The delay activity should block persistence triggering. The workflow cannot be persisted/unloaded while there is a active delay activity.

### DIFF
--- a/src/UiPath.Workflow.Runtime/Statements/Delay.cs
+++ b/src/UiPath.Workflow.Runtime/Statements/Delay.cs
@@ -97,6 +97,6 @@ public sealed class Delay : NativeActivity
 
     private bool HasBlockingDelay(NativeActivityContext context)
     {
-        return context.GetExtension<WorkflowSettingsExtension>()?.BlockingDelay == true;
+        return context.GetExtension<StatementsBehaviorExtension>()?.BlockingDelay == true;
     }
 }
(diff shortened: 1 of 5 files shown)


src/UiPath.Workflow.Runtime/Statements/Delay.cs
```diff
@@ -76,6 +85,16 @@ protected override void Abort(NativeActivityAbortContext context)
         base.Abort(context);
     }
 
+    private void OnTimerFired(NativeActivityContext context, Bookmark bookmark, object value)
```
vuplea: afaik we don't need to call exit, as the variable goes out of scope and the NoPersistHandle disappears.

src/UiPath.Workflow.Runtime/Statements/Delay.cs
```diff
@@ -62,6 +71,9 @@ protected override void Cancel(NativeActivityContext context)
         timerExtension.CancelTimer(timerBookmark);
         context.RemoveBookmark(timerBookmark);
         context.MarkCanceled();
+
+        if (HasBlockingDelay(context))
```
vuplea: we shouldn't need it here either; see the NoPersistScope statement (activity); it doesn't bother calling Exit at all.

vuplea (approved):

PR #348, opened 2025-01-21
Bump signing reference

andreeadumitru:


azure-pipelines.yaml
```diff
@@ -15,7 +15,7 @@ resources:
     - repository: templates
       type: github
       name: UiPath/AzurePipelinesTemplates
-      ref: refs/tags/uipath.package-signing.1.9.2
+      ref: refs/tags/uipath.package-signing.2.0.5
```
andrei-ungureanu-uipath: we can use 2.0.6 like we do in the Studio repo

aoltean16 (approved):

PR #351, opened 2025-01-30
ROBO-4469 - Makes sure that CoreWF doesn't need JIT in case of special chars

mciureanu:
Fixed: If a special character (“ (0x201C)) is present in an VB expression in a xaml, aot compilation didn't work, and JIT was triggered.

### DIFF
--- a/src/Test/TestCases.Workflows/WF4Samples/Expressions.cs
+++ b/src/Test/TestCases.Workflows/WF4Samples/Expressions.cs
@@ -108,10 +108,10 @@ Iterate ArrayList
         }
 
 
-        sealed class JustInTimeCompilerWrapper : JustInTimeCompiler
+        sealed class CompileSpecialCharactersHelperCompiler : JustInTimeCompiler
         {
             private JustInTimeCompiler _inner;
-            public JustInTimeCompilerWrapper(JustInTimeCompiler inner) { this._inner = inner; }
+            public CompileSpecialCharactersHelperCompiler(JustInTimeCompiler inner) { this._inner = inner; }
 
             public override LambdaExpression CompileExpression(ExpressionToCompile compilerRequest)
             {
@@ -135,7 +135,7 @@ Iterate ArrayList
         public void CompileSpecialCharacters()
         {
             var cf = VisualBasicSettings.Default.CompilerFactory;
-            VisualBasicSettings.Default.CompilerFactory = a => new JustInTimeCompilerWrapper(cf(a));
+            VisualBasicSettings.Default.CompilerFactory = a => new CompileSpecialCharactersHelperCompiler(cf(a));
             var activity = Compile(TestXamls.SpecialCharacters);
             var result = TestHelper.InvokeWorkflow(activity);
             VisualBasicSettings.Default.CompilerFactory = cf;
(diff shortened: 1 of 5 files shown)


src/Test/TestCases.Workflows/WF4Samples/Expressions.cs
```diff
@@ -106,6 +106,42 @@ public void CompileCSharpCalculation()
...
+            public JustInTimeCompilerWrapper(JustInTimeCompiler inner) { this._inner = inner; }
+
+            public override LambdaExpression CompileExpression(ExpressionToCompile compilerRequest)
+            {
+                //Make sure to trigger exception only if called from the CompileSpecialCharacters test, since this is injected
+                //in the static VisualBasicSettings.Default.CompilerFactory that affects all tests.
+                var stackTrace = new System.Diagnostics.StackTrace();
+                var frames = stackTrace.GetFrames();
+
+                bool isCalledFromCompileSpecialCharacters =  frames?.Any(frame =>
```
vuplea: let's use AsyncLocal for this check

src/Test/TestCases.Workflows/WF4Samples/Expressions.cs
```diff
@@ -106,6 +106,42 @@ public void CompileCSharpCalculation()
             var activity = Compile(TestXamls.CSharpCalculation);
             TestHelper.InvokeWorkflow(activity, CSharpCalculationInputs).ShouldBe(CSharpCalculationResult);
         }
+
+
+        sealed class JustInTimeCompilerWrapper : JustInTimeCompiler
```
vuplea: maybe a more specific name? e.g. CompileSpecialCharactersHelperCompiler?  If you see it in a stack trace it could help.

src/UiPath.Workflow.Runtime/Expressions/CompiledExpressionInvoker.cs
```diff
@@ -192,8 +192,21 @@ private bool TryGetCurrentCompiledExpressionRoot(ActivityContext activityContext
...
+        if (_textExpression.Language == "VB")
+        {
+            //These characters are not allowed in VB code, so when VB compilation is done, they are replaced with normal quotes.
+            //But in the Expression Tree, we can have them, since they can be parsed directly from xaml code, or from a C# dll,
+            // so we need to "normalize" them, so that the comparison will work. 
+            normalizedExpressionText = normalizedExpressionText
+                .Replace('“', '"')
+                .Replace('”', '"');
+        }
+
```
aoltean16, line 207: I was looking at this and made me wonder: are we placing the fix in the right place? I mean, sure, it is the most elegant approach, that gives us the benefit of not needing to re-compile existing projects and would work out-of-the-box, but on the other hand, I feel like the proper approach is to normalize the text closer to the design-experience, more precisely - when adding the text via the textbox, or when setting a ModelProperty value (in case of viewmodels). @vuplea, @marius-bughiu what do you think?

marius-bughiu: Ideally, we would not allow them to begin with (we'd throw a validation error), but that would be a breaking change.
We can look at normalizing the values design-time as well, but I think we still need this one as well in order to support existing projects (and projects which will be migrated).

aoltean16: so if we all agree its the best compromise, let's merge it. Indeed it brings the above-mentioned benefits 👍

aoltean16 (approved):

aoltean16 (approved):

vuplea (approved):

PR #352, opened 2025-02-03
STUD-74381 Can't select global variables in activity designer

MariusCuciureanu:
In JitCompiler we need to resolve assemblies that may have been loaded in a different AssemblyLoadContext. We do this on the Resolving event by searching in the current AppDomain.

In the past we also called loadContext.LoadFromAssemblyName(assemblyName) which entered a loop throwing StackOverflowException. This call is not needed (even if we detach the Resolving event) since any missing assembly will trigger an error earlier in the compilation step.

https://uipath.atlassian.net/browse/STUD-74381


src/UiPath.Workflow/Activities/ScriptingJitCompiler.cs
```diff
@@ -100,6 +101,16 @@ TextExpressionCompilerResults TryBuild()
...
     }
 
+    private Assembly CollectibleAlc_Resolving(AssemblyLoadContext loadContext, AssemblyName assemblyName)
+    {
+        // The assembly may have been loaded in a separate AssemblyLoadContext and should be available in the current AppDomain.
+        var assembly = AppDomain.CurrentDomain.GetAssemblies().LastOrDefault(a => a.FullName == assemblyName.FullName);
+        if (assembly != null)
+            return assembly;
+
+        return null;
```
aoltean16:
```suggestion
        return AppDomain.CurrentDomain.GetAssemblies().LastOrDefault(a => a.FullName == assemblyName.FullName);
```

aoltean16 (approved):

PR #353, opened 2025-02-04
[ROBO-4334] Add OnActivityChange event

daniel-stanciu:


erghe:
closed because we are not working actively on it. we will reopen it once we start working

PR #354, opened 2025-02-17
STUD-74608-Improve-validation

mciureanu:
Validation with ValidationSettings.SkipImplementationChildren = true, will skip "implementation" children validation, speeding things up. 

Exception: DynamicActivities will not be skipped.

In order to skip our libraries there are two options:
- We mark individual libraries in a way so that we would skip validation
- We define a more generic way how an activity should look like, in order to be skipped. 

For having immediate results, this PR is about the second choice. We noticed that our libraries are defined by a custom activity with static arguments, but dynamic "implementation" usually of a sequence, or other activity. So this is our ideal case. We also noticed that in our designer the parent activity is a DynamicActivity, which would mean that everything in the current workflow will be skipped, so I added this extra condition to not skip DynamicActivities. A little bit "custom" I guess. 

If not covered by e2e tests, or manually tested thoroughly, there is some risk that we skip validation in undesired cases, like using some constraints on some activities that would validate their parents for example (tbd).

### DIFF
--- a/src/UiPath.Workflow.Runtime/ActivityUtilities.cs
+++ b/src/UiPath.Workflow.Runtime/ActivityUtilities.cs
@@ -596,15 +596,16 @@ internal static class ActivityUtilities
             };
 
             int nextEnvironmentId = 0;
+            bool shouldProcessChildren = !options.SkipPrivateChildren || (activity is DynamicActivity);
 
             ProcessChildren(activity, activity.Children, ActivityCollectionType.Public, true, ref nextActivity, ref activitiesRemaining, ref tempValidationErrors);
             ProcessChildren(activity, activity.ImportedChildren, ActivityCollectionType.Imports, true, ref nextActivity, ref activitiesRemaining, ref tempValidationErrors);
-            ProcessChildren(activity, activity.ImplementationChildren, ActivityCollectionType.Implementation, !options.SkipPrivateChildren, ref nextActivity, ref activitiesRemaining, ref tempValidationErrors);
+            ProcessChildren(activity, activity.ImplementationChildren, ActivityCollectionType.Implementation, shouldProcessChildren, ref nextActivity, ref activitiesRemaining, ref tempValidationErrors);
 
             ProcessArguments(activity, activity.RuntimeArguments, true, ref newImplementationEnvironment, ref nextEnvironmentId, ref nextActivity, ref activitiesRemaining, ref tempValidationErrors);
 
             ProcessVariables(activity, activity.RuntimeVariables, ActivityCollectionType.Public, true, ref newPublicEnvironment, ref nextEnvironmentId, ref nextActivity, ref activitiesRemaining, ref tempValidationErrors);
-            ProcessVariables(activity, activity.ImplementationVariables, ActivityCollectionType.Implementation, !options.SkipPrivateChildren, ref newImplementationEnvironment, ref nextEnvironmentId, ref nextActivity, ref activitiesRemaining, ref tempValidationErrors);
+            ProcessVariables(activity, activity.ImplementationVariables, ActivityCollectionType.Implementation, true, ref newImplementationEnvironment, ref nextEnvironmentId, ref nextActivity, ref activitiesRemaining, ref tempValidationErrors);
 
             if (activity.HandlerOf != null)
             {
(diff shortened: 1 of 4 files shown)


marius-bughiu (approved):

src/UiPath.Workflow.Runtime/ActivityUtilities.cs
```diff
@@ -653,7 +654,7 @@ private static void ProcessActivity(ChildActivity childActivity, ref ChildActivi
             // ProcessDelegates uses activity.Environment
             ProcessDelegates(activity, activity.Delegates, ActivityCollectionType.Public, true, ref nextActivity, ref activitiesRemaining, ref tempValidationErrors);
             ProcessDelegates(activity, activity.ImportedDelegates, ActivityCollectionType.Imports, true, ref nextActivity, ref activitiesRemaining, ref tempValidationErrors);
-            ProcessDelegates(activity, activity.ImplementationDelegates, ActivityCollectionType.Implementation, !options.SkipPrivateChildren, ref nextActivity, ref activitiesRemaining, ref tempValidationErrors);
+            ProcessDelegates(activity, activity.ImplementationDelegates, ActivityCollectionType.Implementation, true, ref nextActivity, ref activitiesRemaining, ref tempValidationErrors);
```
aoltean16, line 657: is this expected to always be true?

src/UiPath.Workflow.Runtime/ProcessActivityTreeOptions.cs
```diff
@@ -265,6 +265,12 @@ public static ProcessActivityTreeOptions GetValidationOptions(ValidationSettings
             result = ValidationOptions;
         }
 
+        if (settings.SkipImplementationChildren)
+        {
+            result = result.Clone();
+            result.SkipPrivateChildren = true;
+        }
```
aoltean16, line 272: I'm afraid we don't really know the implications of this change. For example, quite some activities we have, derive from Activity, which means they have the Implementation property populated. Would we skip those as well? and if we do, that means it will skip them always, so it won't even work, as their metadata won't be parsed not even once.

src/UiPath.Workflow.Runtime/ActivityUtilities.cs
mciureanu: Before it was always true, since SkipPrivateChildren was always false. I'm not even sure what are the implementation delegates, and why you would skip them. Skipping implementation children is what we want I think.

src/UiPath.Workflow.Runtime/ProcessActivityTreeOptions.cs
mciureanu: I guess that's the exact idea, to skip the implementation details in libraries, and as far as I saw, libraries are derived from an Activity class, have some "public" arguments in/out and variables, and the rest are "implementation details" that are available only at runtime (or when we cache the metadata), through set implementation method (which is in fact an implementation child). I am investigating if there are some very special libraries that would still want to have validation in the designer in these implementation children. I didn't find anything for now.

andrei-balint: Let's take this example: [System/UiPath.System.Activities/Workflow/MultipleAssign.cs .](https://github.com/UiPath/Activities/blob/96419ef39e3e7faad3851863ca9297fd034302f5/System/UiPath.System.Activities/Workflow/MultipleAssign.cs#L20) . The implementation of CacheMetadata does take into account the implementation details.
Counter example though: https://github.com/UiPath/Activities/blob/96419ef39e3e7faad3851863ca9297fd034302f5/Pipelines/UiPath.Pipelines.Activities/Stage.cs#L40 , where the Implementation is not checked by CacheMetadata

mciureanu: I checked MultipleAssign. I thought the same at first, but I tested, and it works OK. The reason is that the children are not "private", and are defined in a static way:

[image](https://github.com/UiPath/Activities/blob/96419ef39e3e7faad3851863ca9297fd034302f5/System/UiPath.System.Activities/Workflow/MultipleAssign.cs#L114)

I believe that some of the stuff in cache metadata is also not needed, since CacheMetadata goes by default to the activity children, but that's another story. 

So I still didn't find an example that doesn't work yet. I hope that if there is something wrong with any of the activities, we will notice during testing phase, but maybe my expectation is wrong, so not sure...

mciureanu: In the end it will be a feature flag when validating, and this is a PR only for CoreWF, I suggest we merge this for CoreWF, and we can test/debate more for the PR in Studio. When we encounter the first problem, we can easily disable this in Studio.

vuplea (approved):
confirming the changes do not affect usual runtime scenario (SkipPrivateChildren was never true)

aoltean16 (approved):
I do not consider this to be a good approach, as we're intentionately skipping validating Implementation activities (which there are atleast 20 in Activities repo). It's @marius-bughiu and @mciureanu's wish we give this a go as-is.

PR #356, opened 2025-03-04
STUD-16372 - Removed special character Literal string conversion

mciureanu:


marius-bughiu:
Did a quick search for `%` and looks like there are usages, most notable is this one I think - https://github.com/UiPath/CoreWF/blob/2165682c62df609fe5b6b3d8b0d92f7a9b8ef62f/src/UiPath.Workflow/XamlIntegration/ActivityWithResultConverter.cs#L130

The fact that there are no tests is not encouraging, we should add some once we properly understand the scenario.

mciureanu:
This is indeed not the best place to do changes, since the problem of the workflow designer. Thank you @marius-bughiu 
I am closing the PR.

PR #358, opened 2025-03-31
STUD-74571  During expression compilation, added parameter rename to avoid possible conflicts + unit tests

mciureanu:
https://uipath.atlassian.net/browse/STUD-74571
During expression compilation, added parameter rename to avoid possible conflicts + unit tests


vuplea (approved):
I think the safer choice would be to dissuade this scenario (naming variable "value").
But the PR does not affect execution when flag is off.

gpanaitescu (approved):

aoltean16 (approved):

PR #361, opened 2025-04-08
STUD-74767 Added an option so that we don't have to compile expressions when validating

mciureanu:
During validation, most of the time is lost when the expressions are compiled (lambda expressions are compiled with JIT). Compiling makes sense for expression reuse, but in some cases, when we don't want to reuse them, we want to optimize for speed, and interpreting the expressions vs compiling them is usually faster.

### DIFF
--- a/src/UiPath.Workflow.Runtime/Validation/ActivityValidationServices.cs
+++ b/src/UiPath.Workflow.Runtime/Validation/ActivityValidationServices.cs
@@ -353,9 +353,10 @@ public static class ActivityValidationServices
                 try
                 {
                     var invoker = new WorkflowInvoker(constraint);
-                    
-                    if (validationContext.Environment.Extensions != null)
-                    foreach (var e in validationContext.Environment.Extensions?.All)
+
+                    var extensions = validationContext.Environment.Extensions?.All;
+                    if (extensions != null)
+                    foreach (var e in extensions)
                     {
                         invoker.Extensions.Add(e);
                     }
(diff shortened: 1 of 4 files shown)


src/UiPath.Workflow.Runtime/Validation/ActivityValidationServices.cs
```diff
@@ -350,7 +352,17 @@ internal static void RunConstraints(ActivityUtilities.ChildActivity childActivit
 
                 try
                 {
-                    results = WorkflowInvoker.Invoke(constraint, inputDictionary);
+                    var invoker = new WorkflowInvoker(constraint);
+                    
+                    if (validationContext.Environment.Extensions != null)
+                    foreach (var e in validationContext.Environment.Extensions?.All)
```
vuplea: null check is weird

src/UiPath.Workflow.Runtime/Validation/ActivityValidationServices.cs
```diff
@@ -350,7 +352,17 @@ internal static void RunConstraints(ActivityUtilities.ChildActivity childActivit
...
+                    {
+                        invoker.Extensions.Add(e);
+                    }
+                    
+                    //add extensions to the invoker
```
vuplea: comment should be above, or not needed at all

vuplea (approved):

aoltean16 (approved):

PR #362, opened 2025-05-28
use a best-effort metadata reference resolver

AndreiSingeorzan:
If assembly A from the set used for compilation depends on assembly B, and some type from assembly B is used in the expression, compilation will fail. 

So use a MetadataReferenceResolver which looks in all contexts for that particular assembly.

For now, this was only done for (compilation for) validation, though I suspect it might be useful for other compilations as well (studio compiler, precompiled expressions for designer and any others).

https://uipath.atlassian.net/browse/STUD-75944

### DIFF
--- a/src/UiPath.Workflow/Validation/ExternalMetadataReferenceResolver.cs
+++ b/src/UiPath.Workflow/Validation/ExternalMetadataReferenceResolver.cs
@@ -6,7 +6,7 @@ using System.Reflection;
 
 namespace UiPath.Workflow.Validation
 {
-    public class ExternalMetadataReferenceResolver : MetadataReferenceResolver
+    internal sealed class ExternalMetadataReferenceResolver : MetadataReferenceResolver
     {
         private readonly Func<AssemblyName, Assembly> _resolver;
 
@@ -33,7 +33,7 @@ namespace UiPath.Workflow.Validation
 
                 assemblyName.SetPublicKeyToken(referenceIdentity.PublicKeyToken.ToArray());
 
-                var assembly = _resolver?.Invoke(assemblyName);
+                var assembly = _resolver.Invoke(assemblyName);
 
                 if (string.IsNullOrEmpty(assembly?.Location))
                     return null;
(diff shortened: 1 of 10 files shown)


aoltean16 (approved):

aoltean16:
@gabriela-lungu-uip could you please review this? It's quite important for Uinified Build release

src/UiPath.Workflow/Validation/ExternalMetadataReferenceResolver.cs
```diff
@@ -0,0 +1,59 @@
...
+                var assemblyName = new AssemblyName
+                {
+                    Name = referenceIdentity.Name,
+                    Version = referenceIdentity.Version,
+                    CultureName = referenceIdentity.CultureName,
+                };
+
+                assemblyName.SetPublicKeyToken(referenceIdentity.PublicKeyToken.ToArray());
+
+                var assembly = _resolver?.Invoke(assemblyName);
```
gabriela-lungu-uip: null check for _resolver was done above

src/UiPath.Workflow/Validation/RoslynExpressionValidator.cs
```diff
@@ -116,7 +116,7 @@ protected string CreateValidationCode(IEnumerable<string> types, string returnTy
     /// </summary>
     /// <param name="assemblies">The list of assemblies</param>
     /// <param name="namespaces">The list of namespaces</param>
-    protected abstract Compilation GetCompilation(IReadOnlyCollection<Assembly> assemblies, IReadOnlyCollection<string> namespaces);
+    protected abstract Compilation GetCompilation(IReadOnlyCollection<Assembly> assemblies, IReadOnlyCollection<string> namespaces, ValidationSettings validationSettings);
```
gabriela-lungu-uip: just looking at the code, seems like RoslynExpressionValidator can be derived in other assemblies except for this one and for that reason, a default value for validationSettings should be provided in both GetCompilation and Validate. In practice ... you decide if it breaks something

src/UiPath.Workflow/Validation/ExternalMetadataReferenceResolver.cs
```diff
@@ -0,0 +1,59 @@
+﻿using Microsoft.CodeAnalysis;
+using System;
+using System.Collections.Immutable;
+using System.Linq;
+using System.Reflection;
+
+namespace UiPath.Workflow.Validation
+{
+    public class ExternalMetadataReferenceResolver : MetadataReferenceResolver
```
gabriela-lungu-uip: can ExternalMetadataReferenceResolver be internal and sealed?

PR #370, opened 2025-08-02
Avoid memory allocations when retrieving OverloadGroupNames

marius-bughiu:
A new `new ReadOnlyCollection<string>` is created every time the property is accessed. Also, for every object instance, if `_overloadGroupNames` is null, a `new List<string>(0)` is created.

For my test project, this means: 
- 1.2 GB of List<string> allocated
- 1 GB of ReadOnlyCollection<string> allocated

image
image
image

Benchmark:

|                Method |      Mean |     Error |    StdDev |  Gen 0 | Allocated |
|---------------------- |----------:|----------:|----------:|-------:|----------:|
|    OverloadGroupNames | 0.8078 ns | 0.0431 ns | 0.0360 ns |      - |         - |
| OverloadGroupNamesOld | 2.8678 ns | 0.1149 ns | 0.1018 ns | 0.0014 |      24 B |

Benchmark code:

```
[MemoryDiagnoser(true)]
    public class OverloadGroupBenchmark
    {
        RuntimeArgument argument;

        [GlobalSetup]
        public void Setup()
        {
            argument = new RuntimeArgument("arg", typeof(object), ArgumentDirection.In, true);
        }

        [Benchmark]
        public ReadOnlyCollection<string> OverloadGroupNames()
        {
            return argument.OverloadGroupNames;
        }

        [Benchmark]
        public ReadOnlyCollection<string> OverloadGroupNamesOld()
        {
            return argument.OverloadGroupNamesOld;
        }
    }
```

### DIFF
--- a/src/UiPath.Workflow.Runtime/RuntimeArgument.cs
+++ b/src/UiPath.Workflow.Runtime/RuntimeArgument.cs
@@ -12,10 +12,13 @@ using Validation;
 public sealed class RuntimeArgument : LocationReference
 {
     private static InternalEvaluationOrderComparer evaluationOrderComparer;
+    private static readonly ReadOnlyCollection<string> emptyOverloadGroups = new(Array.Empty<string>());
+
     private Argument _boundArgument;
     private readonly PropertyDescriptor _bindingProperty;
     private readonly object _bindingPropertyOwner;
     private List<string> _overloadGroupNames;
+    private ReadOnlyCollection<string> _overloadGroupNamesReadOnly;
     private int _cacheId;
     private readonly string _name;
     private uint _nameHash;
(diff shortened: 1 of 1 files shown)


aoltean16 (approved):

gabriela-lungu-uip (approved):
It's a bit suspish that _overloadGroupNames is not ReadOnlyCollection but is List as its only role is to be returned from OverloadGroupNames property. But its not something you haven't changed and I wouldn't either .... just mentioning it as it may be used in some reflexion code 👀

marius-bughiu:
> It's a bit suspish that _overloadGroupNames is not ReadOnlyCollection but is List as its only role is to be returned from OverloadGroupNames property. But its not something you haven't changed and I wouldn't either .... just mentioning it as it may be used in some reflexion code 👀

It's because I didn't want to do the `if` and the `new ReadOnlyCollection` inside the constructor, as that would in theory be a 'regression' for cases in which you don't access the `OverloadGroupNames` property. So only the flow where you access the property is affected.

PR #372, opened 2025-08-13
CoreWF: Unexpected validation error received in the Expression Editor STUD-72767

daniel-dumitrascu:
Description:
Changed the type string parsing from using commas to using pipes (|) as separators when joining type information, because certain types (like tuples) can contain commas within their names, which would break the parsing logic.

Jira:
https://uipath.atlassian.net/browse/STUD-72767

### DIFF
--- a/src/Test/TestCases.Workflows/CSharpCompilerHelperTests.cs
+++ b/src/Test/TestCases.Workflows/CSharpCompilerHelperTests.cs
@@ -7,7 +7,7 @@ using System;
 using System.Collections.Generic;
 using Xunit;
 
-namespace Test.TestCases.Activities
+namespace TestCases.Workflows
 {
     public class CSharpCompilerHelperTests
     {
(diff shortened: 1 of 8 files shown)


src/Test/TestCases.Workflows/CSharpCompilerHelperTests.cs
```diff
@@ -0,0 +1,229 @@
+﻿using System.Activities;
+using System.Linq;
+using Microsoft.CodeAnalysis;
+using Microsoft.CodeAnalysis.CSharp;
+using Shouldly;
+using System;
+using System.Collections.Generic;
+using Xunit;
+
+namespace Test.TestCases.Activities
```
gpanaitescu: IMO, both test classes should be in the same namespace (TestCases.Workflows)

src/Test/TestCases.Workflows/CSharpCompilerHelperTests.cs
```diff
@@ -0,0 +1,229 @@
...
+            // Arrange
+            var types = Enumerable.Range(0, 17).Select(i => $"T{i}").ToArray();
+            var names = Enumerable.Range(0, 17).Select(i => $"arg{i}").ToArray();
+            var code = "arg0";
+
+            // Act
+            var result = _compilerHelper.CreateExpressionCode(types, names, code);
+
+            // Assert
+            result.ShouldContain("public delegate TResult Func");
```
gpanaitescu, line 126: Please add an assert with the entire result string

src/Test/TestCases.Workflows/CSharpCompilerHelperTests.cs
```diff
@@ -0,0 +1,229 @@
...
+
+            // Act
+            var result = _compilerHelper.CreateExpressionCode(types, names, code);
+
+            // Assert
+            result.ShouldBe("public static Expression<Func<>> CreateExpression() => () => \"Hello World\";");
+        }
+
+        [Fact]
+        public void CreateExpressionCode_WithMoreThan16Parameters_ShouldGenerateCustomDelegate()
```
gpanaitescu, line 115: A test for less than 16 parameters would be valuable, to emphasize the difference

gpanaitescu (approved):

aoltean16 (approved):

PR #388, opened 2025-12-03
Improve exception message for "Expression Activity type 'VisualBasicValue`1' requires compilation in order to run" (STUD-78425)

gpanaitescu:
Added: parent activity display name, argument display name, expression text.

https://uipath.atlassian.net/browse/STUD-78425

### DIFF
--- a/src/Test/TestCases.Workflows/TestUtils/Project.cs
+++ b/src/Test/TestCases.Workflows/TestUtils/Project.cs
@@ -2,34 +2,54 @@
 using Microsoft.CodeAnalysis.CSharp;
 using Microsoft.CodeAnalysis.Host.Mef;
 using Microsoft.CodeAnalysis.Text;
+using Microsoft.CodeAnalysis.VisualBasic;
 using System;
 using System.Activities;
 using System.Activities.ExpressionParser;
+using System.Collections.Generic;
 using System.Linq;
 using System.Reflection;
 using System.Runtime.Loader;
 using System.Threading.Tasks;
+using TestCases.Workflows.WF4Samples;
 
 namespace TestCases.Workflows.TestUtils
 {
     internal class Project
     {
-        static readonly MefHostServices HostServices = MefHostServices.Create(new[]{ "Microsoft.CodeAnalysis.Workspaces",
+        static readonly MefHostServices HostServicesCS = MefHostServices.Create(new[]{ "Microsoft.CodeAnalysis.Workspaces",
             "Microsoft.CodeAnalysis.CSharp.Workspaces", "Microsoft.CodeAnalysis.Features", "Microsoft.CodeAnalysis.CSharp.Features" }
             .Select(Assembly.Load));
-        private readonly AdhocWorkspace _workspace = new(HostServices);
+        static readonly MefHostServices HostServicesVB = MefHostServices.Create(new[]{ "Microsoft.CodeAnalysis.Workspaces",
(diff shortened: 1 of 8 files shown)


src/UiPath.Workflow.Runtime/ActivityInstance.cs
```diff
@@ -857,10 +859,19 @@ internal bool ResolveArguments(ActivityExecutor executor, IDictionary<string, ob
...
-                        completedSynchronously = false;
-                        break;
+                        if (!InternalTryPopulateArgumentValueOrScheduleExpression(argument, i, executor, argumentValueOverrides, resultLocation, false))
+                        {
+                            completedSynchronously = false;
+                            break;
+                        }
+                    }
+                    catch (NotSupportedException ex) when (ex.Data.Contains(CompiledExpressionInvoker.TextExpressionMetadataRequiresCompilationKey) &&
+                        (ex.Data[CompiledExpressionInvoker.TextExpressionMetadataRequiresCompilationKey] as bool? == true))
```
andrei-balint, line 871: we know that if the key is present, the value is always true, maybe just check the key

andrei-balint (approved):

src/UiPath.Workflow.Runtime/ActivityInstance.cs
```diff
@@ -877,6 +888,33 @@ internal bool ResolveArguments(ActivityExecutor executor, IDictionary<string, ob
...
+            }
+
+            if (!string.IsNullOrEmpty(argumentDisplayName))
+            {
+                messageBuilder.Append(argumentDisplayName).Append(' ');
+            }
+
+            messageBuilder.Append(exceptionMessage);
+            return messageBuilder.ToString();
+        }
```
bogdanStan92, line 906: nit: can be simplified, there is no need for SB in this case

```
        static string CreateArgumentErrorMessage(RuntimeArgument? argument, Exception exception)
        {
            if (argument is null)
            {
                return exception.Message;
            }

            var owner = argument.Owner?.DisplayName;
            var argName = argument.DisplayName ?? argument.Name;

            var ownerPart = string.IsNullOrEmpty(owner) ? "" : $"{owner}: ";
            var argPart = string.IsNullOrEmpty(argName) ? "" : $"{argName} ";

            return $"{ownerPart}{argPart}{exception.Message}";
        }
```

src/Test/TestCases.Workflows/TestUtils/Project.cs
```diff
@@ -2,34 +2,54 @@
...
+            .Select(Assembly.Load));
+        private readonly Dictionary<Language, AdhocWorkspace> _workspaces = new()
+        {
+            { Language.CSharp, new AdhocWorkspace(HostServicesCS) },
+            { Language.VisualBasic, new AdhocWorkspace(HostServicesVB) }
+        };
         private readonly MetadataReference[] _references;
         public Project(MetadataReference[] references) => _references = references;
-        public async Task<Type> Compile(string classCode, string className)
+        public async Task<Type> Compile(string classCode, string className, Language language)
```
bogdanStan92, line 33: There is a problem (it was present before) - when trying to run Compile multiple times (e.g. using theory instead of fact).
AssembyLoad will fail (`Assembly with same name is already loaded`).

I am not saying that we should fix this now (or maybe never), but we need to be aware of the problem

bogdanStan92 (approved):

vuplea (approved):

